PR #4, opened 2017-05-31
Rewrite of MNIST submodule

Evizero:
This PR rewrites the MNIST submodule in such a way that a user needs to explicitly acknowledge the download. I think this is in general a fair model for working with external datasets. 

screen shot 2017-05-31 at 15 58 36

It also adds documentation and a little more flexible code that can read just specific images. 
Plus I implement some utility function for converting the MNIST data to proper Julia images with the color interpretation listed on the official websites (black digits on white background)

screen shot 2017-05-31 at 16 01 47

~~Still need to write some tests~~ done


Evizero:
@hshindo Unless you object I'll merge this sometime in the next two days. I plan on giving CIFAR the same treatment.

Evizero:
sorry, didn't mean to do that actually. That was a misclick

hshindo:
@Evizero 
Thanks for your hard work.
`convert2image` is a good feature for image datasets.
I also agree that a user needs to explicitly acknowledge the download.

But I don't think we need many functions such as `traintensor`, `trainlabels`, `traindata`, etc, which complicates the codes too much.
Just providing `x, y = traindata()` and `x, y = testdata()` is simple and enough, I guess.

Evizero:
Thanks for the feedback!

My main argument for having `traintensor` is two-fold.

1.  I often find myself just wanting to access some images and I rather not use a two step process or awkward indexing because one has to go through `traindata(1)[1]`. I expect I am not alone in this

2. The images have to be loaded separately from the labels anyway, so some function like that has to exist

I suppose its fair to say that I am trying to go for user convenience. I'd like the package to turn into the goto source for ML datasets and I personally prefer more code if it makes using the package more convenient.

What if we only export `traindata` and `testdata`, while keeping the option that a user says `MNIST.traintensor` etc ? This way we could keep `traindata` and `testdata` as the quasi standard interface for across datasets while allowing for dataset specific convenience functions.

Evizero:
To raise another argument for why I think such a detailed interface is important for a high quality datasets package.

A simple interface to the MNIST data has been around for a while now with the MNIST.jl package.
Yet [MXNet.jl](https://github.com/dmlc/MXNet.jl/blob/master/src/util.jl#L10-L34) and [Knet.jl](https://github.com/denizyuret/Knet.jl/blob/master/examples/mnist.jl#L84-L99) choose to create their own binding. 

I think if we aim for a high quality datasets package, then it makes sense to expose both low-level and high level code, so however a user wants to work with a dataset such as MNIST he/she can. For example using the `MNIST.Reader` sub-module a user can load images in their native `UInt8` format. I think its flexibility and convenience that makes a great package.

Issue #7, opened 2017-09-22
Publish to METADATA


CarloLucibello:
Hi,
this package seems useful, in good shape, and also easy to mantain. Is there something preventing official release?

Evizero:
Personally I'd say two things should happen first (which I plan on doing eventually)

- [ ] more formal testing (similar to MNIST submodule)
- [ ] adding an "agreement" dialogue to the other dataset download functions (also in spirit of the MNIST submodule)

CarloLucibello:
The second point could be addressed having a common download utility for all the modules, maybe with a signature like this
```julia
download_helper(baseurl, files; i_accept_the_terms_of_use = false, prompt_info = nothing)
```
where for MNIST for example
```julia
prompt_info = """
Dataset: THE MNIST DATABASE of handwritten digits
Authors: Yann LeCun, Corinna Cortes, Christopher J.C. Burges
Website: http://yann.lecun.com/exdb/mnist/

[LeCun et al., 1998a]
  Y. LeCun, L. Bottou, Y. Bengio, and P. Haffner. 
"Gradient-based learning applied to document recognition." 
Proceedings of the IEEE, 86(11):2278-2324, November 1998
"""
```

CarloLucibello:
maybe the package can just be published now and things improved later. I find the package very handy. It is a pity it's not just a Pkg.add  away, this refrains people from using it

rdeits:
Any updates on this? [MNIST.jl](https://github.com/johnmyleswhite/MNIST.jl) appears abandoned and will be broken in v0.7/1.0 because it still hasn't updated its `Array` constructor syntax. That will mean that `Flux.jl`'s MNIST demo will be broken on 1.0, which might be a hindrance for new users.

Evizero:
I have plans to get this ready for metadata over new years

rdeits:
Awesome, thank you!

amellnik:
See also https://github.com/FluxML/model-zoo/issues/15

Evizero:
https://github.com/JuliaLang/METADATA.jl/pull/13399

PR #8, opened 2017-09-23
[WIP] add FashionMNIST dataset

CarloLucibello:
A new and fashionable dataset for MLDatasets
https://github.com/zalandoresearch/fashion-mnist

Most of the code is a plain copy of the one in MNIST module. With some little factorization the duplication can be avoided. But maybe we want first to merge this and then discuss design #7 ?


Evizero:
I think adding this dataset is a good idea. I think it should be possible to reuse the MNIST reader. At least we should make it so

CarloLucibello:
yes, the reader could and should be reused, but  the download function should be factorized out first. I'll try something

Evizero:
Ah yes, I am browsing over the reader code right now. I forgot how integrated the download function is. The motivation for that was that a user gets prompted for download in case the data is missing (for convenience pretty much).

If you have an idea I am all ears. Otherwise I am ok solving this at some other point

Evizero:
Could you maybe also add a submodule readme ? (like shown here https://github.com/JuliaML/MLDatasets.jl/tree/master/src/MNIST)

src/FashionMNIST/reader.jl
```diff
@@ -0,0 +1,13 @@
...
+    Dataset: THE FashionMNIST DATABASE of fashion products
+    Authors: Han Xiao, Kashif Rasul, Roland Vollgraf
+    Website: https://github.com/zalandoresearch/fashion-mnist
+
+    Paper: Han Xiao, Kashif Rasul, Roland Vollgraf "Fashion-MNIST: a Novel Image Dataset for Benchmarking Machine Learning Algorithms."
+
+    The files are available for download at the offical website linked above.
+    """)
+
+Reader.set_baseurl("http://fashion-mnist.s3-website.eu-central-1.amazonaws.com/")
```
Evizero: I like the idea, but how predictable is it to overwrite a global variable? Can one load both datasets in the same session?

CarloLucibello: Good point, hmhmh. Let me think a bit more, maybe I can come up with something better, otherwise I'll revert to the first commit

Evizero:
I think one feasible idea could be to refactor the `dir` parameter that gets passed around in the reader (e.g. https://github.com/JuliaML/MLDatasets.jl/blob/master/src/MNIST/Reader/readimages.jl#L97), to some `Reader.RemoteSettings` struct. This could then include all the settings such as prompt, url, etc.

This would also be the first step towards a cross dataset downloader, but for now MNIST-like specific

Evizero:
I'll merge this into its dedicated branch. I have some downtime and will take a crack at solving this

CarloLucibello:
ok, thanks, couldn't spare time for this last week

Evizero:
see #9 . I'll leave it open for a day or two if you want to take a look.

PR #9, opened 2017-09-30
finish FashionMNIST integration

Evizero:
Conclusion of #8. Now both `MNIST` and `FashionMNIST` use the `MNIST.Reader`. There is some cross-dataset downloader logic now that may be used/adapted for the other datasets if possible.


Evizero:
The relevant tests for MNIST and FashionMNIST pass on travis. 

Fixing the other datasets should be addressed in a different PR

Evizero:
Would be nice to get an OK from you if you have time/interest, since you are the original package author.

Unless any opposition comes up I will merge in a couple days

hshindo:
Good job! Looks good to me.

PR #12, opened 2017-12-26
Switch download backend to DataDeps

Evizero:
@oxinabox

My first attempt to build this package on top of [DataDeps.jl](https://github.com/oxinabox/DataDeps.jl). So far I adapted MNIST and FashionMNIST and it works.

Everything in `src/io/download.jl` is basically glue code to communicate with DataDeps. I first tried to just use `MNIST.traintensor(...; dir = datadep"MNIST")` etc, but it doesn't quite replace the current behaviour. (basically because this approach would only help download the files if no custom dir is chosen.


oxinabox:
Cool,
I think the behavior you desire is just slightly different from what I desire/envision.
The resulting issues on the DataDeps.jl repos, should make it much more flexible.

Evizero:
I am now finished with the rewrite (plus I ported CIFAR-10 as well). This PR is now ready for comments/review if anyone wants to provide some.

After merging this I plan to port the remaining CIFAR-100 image dataset and maybe add some documenter docs, and then I'd say we're ready for METADATA !

src/CIFAR10/CIFAR10.jl
```diff
@@ -0,0 +1,95 @@
...
+            this archive for you if you wish, but that doesn't
+            free you from the burden of using the data
+            responsibly and respect copyright. The authors of
+            CIFAR-10 aren't really explicit about any terms of
+            use, so please read the website to make sure you want
+            to download the dataset.
+            """,
+            "https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz",
+            "c4a38c50a1bc5f3a1c5537f2155ab9d68f9f25eb1ed8d9ddda3db29a59bca1dd",
+            fetch_method = (src, dst) -> run(BinDeps.download_cmd(src, dst)),
```
oxinabox: Why `BinDeps.download_cmd` ?
and not `Base.download` ? (which is the default if not specified)

Compare and contrast:

 - https://github.com/JuliaLang/julia/blob/195008690739cc98719773c734a07b41a618906c/base/interactiveutil.jl#L640-L692
 - https://github.com/JuliaLang/BinDeps.jl/blob/803ca7975d8b5f43610fcb7504f09f2e573b82fd/src/BinDeps.jl#L51-L84

src/CIFAR100/CIFAR100.jl
```diff
@@ -0,0 +1,87 @@
...
+
+        traindata,
+        testdata,
+
+        convert2image,
+        convert2features,
+
+        download
+
+    const DEPNAME = "CIFAR100"
```
oxinabox: Is this good flexible code?
Or is it actually more confusing than it is worth?
I'm not sure.

I guess it lets you copy and past lines that mean the same thing that use the same variable names, in different submodules

src/CIFAR100/README.md
```diff
@@ -0,0 +1,85 @@
...
+## Usage
+
+This sub-module provides a programmatic interface to download,
+load, and work with the CIFAR-100 dataset.
+
+```julia
+using MLDatasets
+
+# download dataset
+CIFAR100.download()
```
oxinabox: I'ld delete this line from the example, since it will be downloaded automatically, and that automatic download is a feature you want to showoff. No?

src/CIFAR100/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+The provided functions also allow for optional arguments, such as
+the directory `dir` where the dataset is located, or the specific
+observation `indices` that one wants to work with. For more
+information on the interface take a look at the documentation
+(e.g. `?CIFAR100.traindata`).
+
+Function | Description
+---------|-------------
+`download([dir])` | Trigger interactive download of the dataset
```
oxinabox: "`download([dir])` | Trigger interactive download of the dataset"

should be something like:

"`download([dir])` | Manually trigger interactive download of the dataset (will be automatically done otherwise, when required)"

src/CIFAR100/interface.jl
```diff
@@ -0,0 +1,394 @@
...
+    @assert mi >= 1 && ma <= TESTSET_SIZE "not all elements in parameter \"indices\" are in 1:$(TESTSET_SIZE)"
+    # preallocate a buffer we will reuse for reading individual
+    # images. "buffer" is written to length(indices) times
+    buffer = Array{UInt8,3}(Reader.NROW, Reader.NCOL, Reader.NCHAN)
+    # we know the types and dimensions of the return values,
+    # so we can preallocate them
+    images = Array{UInt8,4}(Reader.NROW, Reader.NCOL, Reader.NCHAN, length(indices))
+    labels_c = Array{Int,1}(length(indices))
+    labels_f = Array{Int,1}(length(indices))
+    file_path = datafile(DEPNAME, TESTSET_FILENAME, dir)
```
oxinabox: I wonder if you want to add just the this module(/each module):
`datafile(TESTSET_FILENAME, dir) = datafile(DEPNAME, TESTSET_FILENAME, dir)`
kinda link you did for `download`.

src/io/download.jl
```diff
@@ -1,78 +1,71 @@
...
-        end
-        if i_accept_the_terms_of_use
-            mkpath(dir)
-            for file in missing
-                url = baseurl * "$file"
-                path = joinpath(dir, file)
-                info("downloading $file from $url to $dir")
-                run(BinDeps.download_cmd(url, path))
-            end
+function datafile(depname, filename, dir = nothing; recurse = true, kw...)
```
oxinabox: as discussed this will moved into DataDeps.jl
Infact like 100% of this except the docstring will move into DataDeps.jl.
Yes?

src/io/download.jl
```diff
@@ -1,78 +1,71 @@
...
+    path = joinpath(datadir(depname, dir; kw...), filename)
+    if !isfile(path)
+        warn("The file \"$path\" does not exist, even though the dataset-specific folder does. This is an unusual situation that may have been caused by a manual creation of an empty folder, or manual deletion of the given file \"$filename\".")
+        info("Retriggering DataDeps.jl for \"$depname\" to \"$dir\".")
+        download_dep(depname, dir; kw...)
+        if recurse
+            datafile(depname, filename, dir; recurse = false, kw...)
```
oxinabox: It is unclear to me what `recurse` does.
I think maybe this line should be:
`datafile(depname, filename, path; recurse = true, kw...)` ?

oxinabox:
Do you want to wait on merging this until DataDeps.jl has been changed?
(Feel free to PR in the helpers you've made, BTW)

src/CIFAR10/CIFAR10.jl
```diff
@@ -0,0 +1,95 @@
...
+
+            The compressed archive file that contains the
+            complete dataset is available for download at the
+            offical website linked above; specifically the binary
+            version for C programs. We can download and unpack
+            this archive for you if you wish, but that doesn't
+            free you from the burden of using the data
+            responsibly and respect copyright. The authors of
+            CIFAR-10 aren't really explicit about any terms of
+            use, so please read the website to make sure you want
```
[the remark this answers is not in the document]
Evizero: Yea I like your suggestion better

src/CIFAR10/CIFAR10.jl
Evizero: unsure, somehow I expected the BinDeps one to be more general/cross platform. And that impression was solely based on the fact that it exists (why else does it exist?)

src/CIFAR100/CIFAR100.jl
Evizero: Well this defines the name of the folder/registers datadep. Given that it is basically the name of the dataset it made sense to me to make it into a variable. Otherwise, changing that at any time would be a pain, since you can't easily search and replace it.

src/CIFAR100/README.md
Evizero: I suppose. Really I do want to get rid of the sub-readmes anyway and create a nice documenter site

src/CIFAR10/CIFAR10.jl
oxinabox: I have wondered why it existed since I first saw it.
I think it might be justreally old, older than `Base.download` (which is itself old).

It might be more general, but it is redundantly so.
In the julia already promises `Base.download` will work-on every platform.
(Or so I believe)

src/io/download.jl
Evizero: This is basically just to avoid an "endless loop" situation if you have a type in the filename. Because if you do, then the download will be triggered (because the file isn't found) and then this function is called again. since the file still can't be found the download would be triggered again, and so on

Evizero:
Thanks for the review!

> Do you want to wait on merging this until DataDeps.jl has been changed?
(Feel free to PR in the helpers you've made, BTW)

I can leave this open, yes. That said, I have spend a lot of time on this now and I focus on some "actual"  work for a while again. But feel free to port code to DataDeps at your discretions. I'd be happy to take the time and provide feedback

oxinabox:
Fair enough. 
That will give me to to begin actionising on your feedback in turn. :-D

src/CIFAR10/CIFAR10.jl
Evizero: Well I don't feel strongly about this, so I'd be ok to change it to download (or delete the line, since download is the default anyway)

oxinabox:
🎉

PR #17, opened 2018-05-18
fix RegisterDataDep deprecation

CarloLucibello:
addresses the following warning:
```julia
WARNING: RegisterDataDep(name::String, message::String, remotepath, hash=nothing; fetch_method=download, post_fetch_method=identity) is deprecated, use register(DataDep(name::String, message::String, remotepath, hash; fetch_method=fetch_method, post_fetch_method=post_fetch_method)) instead.
```


Evizero:
Thanks! can you also bump the minimum DataDep version in REQUIRE?

PR #18, opened 2018-07-09
fix  deprecation

CarloLucibello:
seems like in #17 I missed one of the RegisterDataDep. Please tag a new version after merging. 

Best,
Carlo


Evizero:
nice! just ran into this warning yesterday actually

Evizero:
Thanks!

Evizero:
https://github.com/JuliaLang/METADATA.jl/pull/15625

Unsure if it will go through. Might be that the HDF5 OSX issue wasn't a fluke

PR #20, opened 2018-08-12
Make ImageCore and MAT optional, switch from GZip to CodecZlib

alyst:
This PR drops Julia v0.6 and fixes most of Julia v0.7/v1.0 deprecations. Plus it does a few other things:
* switches to [CodecZlib.jl](https://github.com/bicycle1885/CodecZlib.jl) from the aging GZip.jl. Unfortunately, it required much more refactoring of the reading code than I originally anticipated (since CodecZlib doesn't support `seek()`). But hopefully these changes simplify the codebase a bit too.
* makes ImageCore.jl and MAT.jl optional (via [Requires.jl](https://github.com/MikeInnes/Requires.jl)). These packages bring with them tons of other dependencies, while the functionality they provide is not essential. ImageCore and MAT are still required for the unit testing. If you install ImageCore and/or MAT in your julia environment, the corresponding MLDatasets functions (converting to images and SVHN2 dataset) would be automatically enabled.
* updates UD_English links

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,4 +30,4 @@ install:
 
 after_success:
   - julia -e 'using Pkg, MLDatasets; cd(joinpath(dirname(pathof(MLDatasets)), "..")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())';
-  - julie -e 'using Pkg; Pkg.add("Documenter"); cd(joinpath(dirname(pathof(MLDatasets)), "..")); include(joinpath("docs", "make.jl"))'
+  - julia -e 'using Pkg, MLDatasets; Pkg.add("Documenter"); cd(joinpath(dirname(pathof(MLDatasets)), "..")); include(joinpath("docs", "make.jl"))'
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,8 +1,7 @@
 julia 0.7
-ImageCore 0.1.2
+Requires
 FixedPointNumbers 0.3
 ColorTypes 0.4
 DataDeps 0.3
-GZip
+CodecZlib
 BinDeps
-MAT
--- a/src/CIFAR10/CIFAR10.jl
+++ b/src/CIFAR10/CIFAR10.jl
@@ -2,27 +2,16 @@ export CIFAR10
 module CIFAR10
     using DataDeps
     using BinDeps
-    using ImageCore
     using ColorTypes
     using FixedPointNumbers
-    using ..MLDatasets: bytes_to_type, datafile, download_dep, download_docstring
+    using ..MLDatasets: bytes_to_type, datafile, download_dep, download_docstring, _colorview, _channelview
 
     export
-
         classnames,
-
-        traintensor,
-        testtensor,
-
-        trainlabels,
-        testlabels,
-
-        traindata,
-        testdata,
-
-        convert2image,
-        convert2features,
-
+        traintensor, testtensor,
+        trainlabels, testlabels,
+        traindata, testdata,
+        convert2image, convert2features,
         download
 
     const DEPNAME = "CIFAR10"
@@ -62,7 +51,7 @@ module CIFAR10
     """
     download(args...; kw...) = download_dep(DEPNAME, args...; kw...)
 
-    include(joinpath("Reader","Reader.jl"))
+    include(joinpath("Reader", "Reader.jl"))
     include("interface.jl")
     include("utils.jl")
 
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -30,7 +30,7 @@ makedocs(
 deploydocs(
     repo = "github.com/JuliaML/MLDatasets.jl.git",
     target = "build",
-    julia = "0.6",
+    julia = "0.7",
     deps = nothing,
     make = nothing,
 )
--- a/docs/src/LICENSE.md
+++ b/docs/src/LICENSE.md
@@ -1,5 +1,6 @@
 # LICENSE
 
 ```@eval
-Markdown.parse_file(joinpath(@__DIR__, "../LICENSE"))
+using Markdown
+Markdown.parse_file(joinpath(@__DIR__, "..", "..", "LICENSE"))
 ```
--- a/src/CIFAR10/Reader/Reader.jl
+++ b/src/CIFAR10/Reader/Reader.jl
@@ -29,13 +29,19 @@ function _readdata!(io::IO,
     else
         buffer = Array{UInt8}(undef, img_size)
         pos = 0
+        last_src_ix = 0
+        last_label = first(labels)
         @inbounds for (dst_ix, src_ix) in zip(dst_indices, src_indices)
-            nextpos = imagesize * (src_ix - src_offset - 1)
-            skip(io, nextpos - pos)
-            labels[dst_ix] = read(io, eltype(labels))
-            read!(io, buffer)
+            if src_ix != last_src_ix
+                nextpos = imagesize * (src_ix - src_offset - 1)
+                skip(io, nextpos - pos)
+                last_label = read(io, eltype(labels))
+                read!(io, buffer)
+                pos = nextpos + imagesize
+                last_src_ix = src_ix
+            end
             copyto!(view(images, :,:,:, dst_ix), buffer)
-            pos = nextpos + imagesize
+            labels[dst_ix] = last_label
         end
     end
     return images, labels
--- a/src/CIFAR10/Reader/Reader.jl
+++ b/src/CIFAR10/Reader/Reader.jl
@@ -1,54 +1,79 @@
 module Reader
 
-export
-
-    readdata!,
-    readdata
+export readdata
 
 const NROW = 32
 const NCOL = 32
 const NCHAN = 3
-const NBYTE = NROW * NCOL * NCHAN + 1 # "+ 1" for label
 const CHUNK_SIZE = 10_000
 
-function readnext!(buffer::Array{UInt8}, io::IO)
-    y = Int(read(io, UInt8))
-    read!(io, buffer)
-    buffer, y
-end
-
-function readdata!(buffer::Array{UInt8}, io::IO, index::Integer)
-    seek(io, (index - 1) * NBYTE)
-    readnext!(buffer, io)
-end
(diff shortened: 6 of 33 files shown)


.travis.yml
```diff
@@ -22,7 +23,11 @@ addons:
...
+#script:
+#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+#  - julia --check-bounds=yes -e 'using Pkg; Pkg.add(pwd()); Pkg.build("MLDatasets"); Pkg.test("MLDatasets"; coverage=true)'
+
 after_success:
-  - julia -e 'cd(Pkg.dir("MLDatasets")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
-  - julia -e 'Pkg.add("Documenter")'
-  - julia -e 'cd(Pkg.dir("MLDatasets")); include(joinpath("docs", "make.jl"))'
+  - julia -e 'using Pkg, MLDatasets; cd(joinpath(dirname(pathof(MLDatasets)), "..")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())';
+  - julie -e 'using Pkg; Pkg.add("Documenter"); cd(joinpath(dirname(pathof(MLDatasets)), "..")); include(joinpath("docs", "make.jl"))'
```
Evizero: the coverage doesn't seem to work. Maybe we should try emulating https://github.com/fredrikekre/Literate.jl/blob/master/.travis.yml

REQUIRE
```diff
@@ -1,8 +1,7 @@
-julia 0.6
-ImageCore 0.1.2
+julia 0.7
+Requires
 FixedPointNumbers 0.3
-ColorTypes 0.4
+ColorTypes
```
Evizero: why did you remove the min version?

src/CIFAR10/CIFAR10.jl
```diff
@@ -59,13 +45,13 @@ module CIFAR10
 
     This function will display an interactive dialog unless
     either the keyword parameter `i_accept_the_terms_of_use` or
-    the environment variable `DATADEPS_ALWAY_ACCEPT` is set to
+    the environment variable `DATADEPS_ALWAYS_ACCEPT` is set to
```
Evizero: good catch!

src/CIFAR10/Reader.jl
```diff
@@ -0,0 +1,69 @@
...
+        end
+    end
+    return images, labels
+end
+
+_readdata!(file::AbstractString, images, labels, nsrc, src_indices,
+           src_offset = 0, dst_indices::AbstractVector{<:Integer} = 1:size(images, 4)) =
+    open(file, "r") do io
+        _readdata!(io, images, labels, nsrc, src_indices, src_offset, dst_indices)
+    end
```
Evizero: i think for functions like this we should use the

```
function ...
end
```

syntax

src/CIFAR10/CIFAR10.jl
alyst: > It makes it harder to review and I don't think there is a reason to do that change. Could you please change that back?

Makes sense. I've restored the file structure.

However, in the long run I think these changes make sense. `Reader` folder have only one `Reader.jl` file; `reader*.jl` files in `MNIST` are very small, also reading images/labels is coupled with reading headers, so I think there's a good reason to keep these functions in one file. Personally, I would also simplify the modules structure, because it makes sharing the code (e.g. between CIFAR10 and CIFAR100) harder, also this boundaries are a bit artificial (e.g. CIFAR10/interface.jl contains the logic to decide from which file to read from).
Of course it makes sense to define and export the modules for the individual datasets, but they may be very small and just contain the `MNIST.traindata/testdata()` public wrappers, while the actual private functions would reside in one single MLDatasets module.

.travis.yml
alyst: From Travis logs it looks like the coverage script was running, but then there were some network problems, so it failed to connect to Coveralls server.

src/CIFAR10/CIFAR10.jl
Evizero: I agree there is an argument to be had to simplify the structure, but this should be separate from this PR. 

You do a lot of objectively important updates in this PR that I am very excited about. Let us not hold those back with subjective preferences.

src/CIFAR10/Reader.jl
alyst: I thought we can use a terse declaration syntax for single-instruction wrapper functions. Just tell me if you think it's critical to revert it.

Evizero: It is not critical. I can do it later

alyst:
So it looks like the tests are running fine on v0.7 (for stripped-down CI tests that only download MNIST, and also locally when I test for all datasets).
On v1.0 it fails because of the deprecations in the dependencies.

The Coveralls coverage is being collected (I've compared it to Literate.jl script and essentially it's the same), but then there is a recurring error when trying to upload it.
Here's an excerpt from Travis logs:
```
Submitting data to Coveralls...
ERROR: HTTP.ExceptionRequest.StatusError(422, HTTP.Messages.Response:
"""
HTTP/1.1 422 Unprocessable Entity
Date: Mon, 13 Aug 2018 10:04:40 GMT
Content-Type: application/json; charset=utf-8
Content-Length: 72
Connection: keep-alive
Set-Cookie: __cfduid=d37098939494e5f2a06d157d3c49c634b1534154680; expires=Tue, 13-Aug-19 10:04:40 GMT; path=/; domain=.coveralls.io; HttpOnly
Cache-Control: no-cache
Status: 422 Unprocessable Entity
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Powered-By: Phusion Passenger Enterprise 5.1.10
X-Request-Id: 06622f00-1946-4c49-aadf-4ba7256abe0b
X-Runtime: 0.240836
X-XSS-Protection: 1; mode=block
Expect-CT: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
Server: cloudflare
CF-RAY: 449a5bdb8fbe240e-IAD
{"message":"Couldn't find a repository matching this job.","error":true}""")
Stacktrace:
 [1] #request#1 at /home/travis/.julia/packages/HTTP/nUK4f/src/ExceptionRequest.jl:22 [inlined]
 [2] (::getfield(HTTP, Symbol("#kw##request")))(::NamedTuple{(:iofunction, :verbose, :cookiejar, 
...
```

Evizero:
I think there are too many big changes here mixed with opinionated changes such as removing testsets.

I am not confortable merging a rewrite like this. For example its not clear to me if the current public interface is preserved. The `skip` implementention looks like it would have problems with dealing with indices that are out of order (like `[3,2,5]`). 

I am not completely rejecting this change, but I think that before we make such a drastic change, I rather first just update to 0.7 in some minimal way and see if a rewrite is needed

alyst:
> I think there are too many big changes here mixed with opinionated changes such as removing testsets.

What tests are you referring to? I think I've removed none. I can restore the indentation in the test, so that it's more clear (sorry for that, I thought GitHub would be able to detect that it's just a whitespace change), but often the reason for changing indentation was that I replaced `for ... @testset "..." begin end` pattern with `@testset "..." for ... end`.

> The skip implementention looks like it would have problems with dealing with indices that are out of order (like [3,2,5]).

It does support out-of-order indices (also see the comments for `_readdata()` methods) and it's covered by the existing tests.

> For example its not clear to me if the current public interface is preserved.

As long as it's covered by the tests, it should be preserved.

I can fix some specific things, but unfortunately I have no time to make iterative rewrites. If you think you don't need these updates, feel free to ignore the PR. I needed these changes to switch TSne.jl from unsupported MNIST.jl to MLDatasets.jl, I guess I will have to wait then.

Evizero:
fair. maybe I am just a little overwhelmed by the sheer amount of changes. As a maintainer there is a certain cost to swapping out the familiar code base with a different one. This step can be a bit intimitating. And if I am misjudging thing such as support for out of order indicies it just demonstrates that I would need to spend some time familiarizing myself with it before I'd be able to maintain it 

Concerning 0.7 update. I am creating a PR in a minute. So at least the package will support 0.7. For 1.0 support there still need to be some upstream fixes to MAT.jl

alyst:
@Evizero Any further comments? I would like to release 0.7-enabled version of TSne.jl that could pass CI tests and I hoped to use MLDatasets.jl as the test data provider.

The general question is whether you would like to introduce the changes that necessitated this amount of code churn. I.e. making MAT and ImageCore optional and switching from Gzip to CodecZlib. If you are fine with that, we can think of the strategies to make you comfortable with the changes. The simplest one would be just to clarify your questions about the new code. I can also try splitting the changes into smaller commits, so it's easier to track what's going on.

Evizero:
As impressed and thankful I am for the amount of work you put into this, I don't think I am ready at this point to swap out GZip for a different backend. Its just too many code changes.

A 0.7 compatible version is tagged that should work on linux at least. OSX has the whole HDF5 issue through MAT. So I am very much in favour of making the whole MAT situation nicer. If you would be interessted into making a separate PR for just the optional dependency stuff, I'd be happy to provide quick feedback to get that part merged and tagged quickly.

alyst:
> If you would be interessted into making a separate PR for just the optional dependency stuff, I'd be happy to provide quick feedback to get that part merged and tagged quickly.

Fair enough, these changes could be extracted into a separate PR. See #23

> I don't think I am ready at this point to swap out GZip for a different backend. Its just too many code changes.

Hopefully you will be fine with #23, and after rebasing this PR on top of it there would be less changes. :)
One may argue whether they are really required given that now `GZip` is updated to support 1.0, but they also reduce the amount IO and CPU, since the archive files are read sequentially. This is a community-maintained package, so I will be glad to help you maintaining the relevant parts.

CarloLucibello:
@alyst do you think it is still worth to move to CodecZlib?

logankilpatrick:
Hey @alyst, it looks like this might have finally caught up to us. Any interest in putting a PR together to help resolve https://github.com/JuliaML/MLDatasets.jl/issues/86 ?

CarloLucibello:
closing as outdated

PR #23, opened 2018-08-23
Make ImageCore and MAT packages optional

alyst:
This PR (extracted from #20) makes ImageCore.jl and MAT.jl packages optional (via [Requires.jl](https://github.com/MikeInnes/Requires.jl)). These packages bring with them tons of other dependencies, while the functionality they provide is not essential. ImageCore and MAT are still required for unit testing. If ImageCore and/or MAT are added to a specific Julia project, the corresponding MLDatasets functions (converting to images and SVHN2 dataset) would be automatically enabled.

As a samll added bonus it should fix Documenter scripts.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,0 +1,2 @@
+ImageCore 0.7
+MAT 0.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,4 +30,4 @@ install:
 
 after_success:
   - julia -e 'using Pkg, MLDatasets; cd(joinpath(dirname(pathof(MLDatasets)), "..")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())';
-  - julie -e 'using Pkg; Pkg.add("Documenter"); cd(joinpath(dirname(pathof(MLDatasets)), "..")); include(joinpath("docs", "make.jl"))'
+  - julia -e 'using Pkg, MLDatasets; Pkg.add("Documenter"); cd(joinpath(dirname(pathof(MLDatasets)), "..")); include(joinpath("docs", "make.jl"))'
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,8 +1,7 @@
 julia 0.7
-ImageCore 0.1.2
+Requires
 FixedPointNumbers 0.3
 ColorTypes 0.4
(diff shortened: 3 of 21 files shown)


test/REQUIRE
```diff
@@ -0,0 +1,2 @@
+ImageCore 0.7
+MAT 0.4+
```
Evizero: is the trailing `+` a typo or on purpose? Have not seen one before

alyst: I put it on purpose as a reminder that MAT v0.4 is not compatible with Julia v1.0 and it should be updated as soon as the new version of MAT is released.
Julia parses it, but IIUC it doesn't conform to semantic versioning specification, so I removed it to avoid problems with METADATA.jl CI tests.

Evizero:
Thanks!

alyst:
@Evizero Thanks! I don't fully understand the purpose of some of your later tweaks, though:
* `Pkg.clone()` is deprecated since v0.7 and explicitly states that `Pkg.add()` should be used instead
* `using X` adds exported definitions from `X` to the search path, `import X` additionally allows extending the methods of `X`. So I'm not sure one really ever wants to do `import Pkg`

Evizero:
1. The `clone` change was an ignorant oversight on my part. I copy and pasted a "new" travis yml from one of my other repos, because I saw the `pathof` line and the way documenter is invoked and thought "oh that looks old". 

2. I don't know what you are referring to specifically. Generally speaking though I think there is a misunderstanding. `import X` does less than `using X`. For example `import X` does not add the exported definitions from `X` to the search path, and so you also can't extend functions of `X` without explicitly prefixing it. extending prefixed functions works for both `import X` and `using X`. What you are hinting at happens when you do `import X: foo` versus `using X: foo`

Evizero:
another comment concerning `clone`: In the future I'll see that I also use PRs for changes such as this, so that I give others (such as you) the chance to point out my mistakes early

alyst:
> I don't know what you are referring to specifically.

It works now, but I'm not sure it's guaranteed that `import Pkg` is fully interchangeable with `using Pkg`. `Pkg` might export something (purely hypothetical example: a variable with the path to the packages repository), which the other parts of the CI script may rely on.

> In the future I'll see that I also use PRs for changes such as this

:+1:

PR #28, opened 2019-09-22
add Project.toml and related updates

CarloLucibello:
this adds a Project.toml and Manifest.toml and removes the REQUIRE files. 
~~I'm sorry for the messy commits, looks like I made some mess with git and I can't manage to squash them together~~

Also, increases version to "0.4" so that we can tag a release after merging this


CarloLucibello:
Squashed the commits. 
I get the following problem with MbedTLS on julia 1.0 and 1.1
```  Building MbedTLS → `~/.julia/packages/MbedTLS/a1JFn/deps/build.log`
3.46s$ julia --check-bounds=yes --color=yes -e "if VERSION < v\"0.7.0-DEV.5183\"; Pkg.test(\"${JL_PKG}\", coverage=true); else using Pkg; Pkg.test(coverage=true); end"
   Testing MLDatasets
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package MbedTLS [739be429]:
 MbedTLS [739be429] log:
 ├─possible versions are: [0.1.0-0.1.4, 0.2.0-0.2.6, 0.3.0, 0.4.1-0.4.5, 0.5.0-0.5.14, 0.6.0-0.6.8, 0.7.0] or uninstalled
 ├─restricted to versions 0.7.0 by an explicit requirement, leaving only versions 0.7.0
 └─restricted by julia compatibility requirements to versions: [0.5.13-0.5.14, 0.6.0-0.6.8] or uninstalled — no versions left
```
julia 1.2 passes test instead. Not sure where the restriction on the MbedTLS version comes from.
Trying with removing the Manifest file

CarloLucibello:
@Evizero this should be ready to merge

CarloLucibello:
@Evizero any feedback?

docs/src/datasets/FashionMNIST.md
```diff
@@ -56,8 +56,8 @@ the Fashion-MNIST dataset in Julia more convenient.
 
 Function | Description
 ---------|-------------
-[`convert2features(array)`](@ref FashionMNIST.convert2features) | Convert the Fashion-MNIST tensor to a flat feature matrix
-[`convert2image(array)`](@ref FashionMNIST.convert2image) | Convert the Fashion-MNIST tensor/matrix to a colorant array
+[`convert2features(array)`](@ref MNIST.convert2features) | Convert the Fashion-MNIST tensor to a flat feature matrix
```
Evizero: does the `FashionMNIST.`one not work anymore? The idea is that even though it literally is just the MNIST one, it could always be replaced with its own implementation as well as everything needed for FashionMNIST is available in its module

CarloLucibello: `@ref FashionMNIST.` does work but the hyperlink redirects to the same html page as `@ref MNIST.`, which I found confusing. So I wanted to make it explicit by switching to `@ref MNIST.` and adding a comment below saying that those two functions are just a re-export from MNIST. I can switch back to `@ref FashionMNIST.` if you think that would be more future proof, I'm indifferent to either choice.

docs/src/datasets/FashionMNIST.md
```diff
@@ -56,8 +56,8 @@ the Fashion-MNIST dataset in Julia more convenient.
...
+[`convert2image(array)`](@ref MNIST.convert2image) | Convert the Fashion-MNIST tensor/matrix to a colorant array
 
 You can use the function
 [`convert2features`](@ref FashionMNIST.convert2features) to
```
Evizero: this (and possibly other) occurances should probably also be changed then for consistency

CarloLucibello: addressed by last commit

Evizero:
thanks

CarloLucibello:
can you activate tagbot for the repo
https://github.com/apps/julia-tagbot
and then register a new release
https://github.com/JuliaRegistries/Registrator.jl
?
The Project.toml already contains the new version

CarloLucibello:
@Evizero could you add me as a mainter of this repo?

Evizero:
done. welcome!

Did i miss a step with the tagging? https://github.com/JuliaML/MLDatasets.jl/commit/4a3ed96d579a84648e15d06c45a70a82a0866bf2#commitcomment-35362595

CarloLucibello:
ah, I missed that. Thanks anyway

PR #34, opened 2020-03-17
return image tensors in WH format

CarloLucibello:
Most downstream packages such as Flux, Knet, and the Images ecosystem, assume the image tensors to be in WH format for greyscale and in WHC format for rgb. 
Our SVHN2 dataset is consistent with julia ecosystem, while MNIST, FashionMNIST, CIFAR10 and CIFAR100 return images HW format. This is annoying at the very least, but most importantly it is the potential source of very hard to detect bugs. 

Therefore this PR does the following:
- makes every vision dataset return tensors in WH format
- Since this is breaking, I'm also bumping version number.  
- I'm removing the `convert2features` since it's just a reshape into a vector (or a matrix for a batch of images)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "MLDatasets"
 uuid = "eb30cadb-4394-5ae3-aed4-317e484a6458"
-version = "0.4.1"
+version = "0.5.0"
 
 [deps]
 BinDeps = "9e28174c-4ba2-5203-b857-d8d62c4213ee"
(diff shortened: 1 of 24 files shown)


johnnychen94:
> Most downstream packages such as Flux, Knet, and the Images ecosystem, assume the image tensors to be in WH format for greyscale and in WHC format for rgb.

Didn't check the code changes. I'm a little confused and I think you're suggesting the opposite. The whole JuliaImages are using column-major order, i.e., HW format

image

> I'm removing the convert2features since it's just a reshape into a vector (or a matrix for a batch of images)

Removing it is fine, but there should be deprecations since this name is exported.

CarloLucibello:
jeez, I got confused by the current docstring saying  
"""
the first dimension corresponds to the pixel *rows* (x) of the
image, the second dimension to the pixel *columns* (y) 
""" 
which means HW (right?) while I get a rotated 9 here:

Schermata del 2020-03-17 13-44-32

CarloLucibello:
So Knet and Flux use WHCN, while Images uses HW. MLDatasets does WHCN although its docstrings say HWCN. What should I do here, just fix the docstring?

johnnychen94:
I think the docstring is correct, because unlike other languages, in julia `row` is the second dimension.

```julia
julia> x = reshape(1:16, 2, 8)./16
2×8 Array{Float64,2}:
 0.0625  0.1875  0.3125  0.4375  0.5625  0.6875  0.8125  0.9375
 0.125   0.25    0.375   0.5     0.625   0.75    0.875   1.0

julia> first(eachrow(x)) == x[1, :]
true
```

CarloLucibello:
wait, looking again at the docstring
"""
the first dimension corresponds to the pixel rows (x) of the
image, the second dimension to the pixel columns (y)
"""
I think it does mean WH although it is very confusing (at least to me) and should be rephrased

johnnychen94:
Yeah, perhaps some emphasis on the layout difference would be sufficient.

PR #35, opened 2020-03-21
change SVHN2 to WH format + other fixes

CarloLucibello:
Revisit of #34 where I messed things up. 
Most downstream packages such as Flux, Knet, and the Images ecosystem, assume the image tensors to be in WH format for greyscale and in WHC format for rgb.

Our MNIST, FashionMNIST, CIFAR10 and CIFAR100 return images in WH (width, height) format  while SVHN2  return HW. The  latter is annoying at the very least, but most importantly it is the potential source of very hard to detect bugs.

Therefore this PR does the following:
- make clear in doctrings that image tensors are in WH
- return SVHN2 image tensors in WH format, so that every vision dataset is in WH format
- deprecate convert2features since it's just a reshape into a vector (or a matrix for a batch of images)
- fix mnist convert2images (which was inverting the greyscale before)
- add MAT.jl to project requires, since it is a lean dependency
- add windows to travis and test on julia 1.4
- Since this is breaking, I'm also bumping version number.


src/CIFAR100/CIFAR100.jl
```diff
@@ -21,10 +21,12 @@ module CIFAR100
         testdata,
 
         convert2image,
-        convert2features,
 
         download
 
+    @deprecate convert2features reshape
```
johnnychen94: Despite of a depwarn, is `CIFAR10.convert2features(CIFAR10.traintensor(Float32,1))` still working without error?

Project.toml
```diff
@@ -9,16 +9,16 @@ DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
...
 GZip = "92fee26a-97fe-5a0c-ad85-20a5f3185b63"
+MAT = "23992714-dd62-5051-b70f-ba57cb901cac"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 
 [compat]
-ColorTypes = "0.4, 0.5, 0.6, 0.7, 0.8, 0.9"
+BinDeps = "1"
+ColorTypes = "0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 0.10"
 DataDeps = "0.3, 0.4, 0.5, 0.6, 0.7"
-FixedPointNumbers = "0.3, 0.4, 0.5, 0.6, 0.7, 0.8"
```
johnnychen94: We still need `FixedPointNumbers` here.

src/CIFAR100/CIFAR100.jl
CarloLucibello: no it will error, but it will but it will point in the right direction
```
julia> CIFAR10.convert2features(rand(32,32,3))
┌ Warning: `convert2features` is deprecated, use `reshape` instead.
│   caller = top-level scope at REPL[4]:1
└ @ Core REPL[4]:1
ERROR: DimensionMismatch("new dimensions () must be consistent with array size 3072")
Stacktrace:
 [1] (::Base.var"#throw_dmrsa#196")(::Tuple{}, ::Int64) at ./reshapedarray.jl:41
 [2] reshape at ./reshapedarray.jl:45 [inlined]
 [3] reshape at ./reshapedarray.jl:116 [inlined]
 [4] convert2features(::Array{Float64,3}) at ./deprecated.jl:43
 [5] top-level scope at REPL[4]:1
```
I think it's ok like this, it's not worth having a proper deprecation, people can figure this out (if anyone at all was using convert2features).

johnnychen94: I think the original design of this name was to introduce a conceptually symmetric name to `convert2image`, regardless of it's being trivial.

I can foresee a lot of example codes broken due to this... and non-power users would keep asking questions about how to fix things if this is totally removed.

Said that, if you think this as an edge usage, feel free to continue the deletion.

.travis.yml
```diff
@@ -1,13 +1,18 @@
...
+
 os:
   - linux
   - osx
+  - windows
+
 julia:
   - 1.0
-  - 1.3
+  - 1.4
```
johnnychen94: we can just use `1` to link to the latest stable `1.x` release.

src/CIFAR100/CIFAR100.jl
CarloLucibello: I do think this is an edge case, with an easy fix pointed by the warning that should be within the reach of non-power users as well. It's not worth investing time in doing the deprecation properly (and I don't have much time these days). Maybe we can deprecate properly later if someone comes up with an issue for this

PR #55, opened 2021-02-04
Include EMNIST

### LINKED ISSUE #51: EMNIST
Love the package. Would EMNIST be small enough to add?

https://www.nist.gov/itl/products-and-services/emnist-dataset

andrew-saydjari:
Fixes #51

I think I have properly included all the data and functionality needed to integrate EMNIST into the package. It passed all tests.

I wrote very little fancy functionality and barebones tests/documentation. I plan to add rotMNIST soon and will try to improve things then time allowing. I really wanted the different datasets packaged in EMNIST to be called like EMNIST.byclass.traindata() etc. I am not sure how I implemented was the cleanest way possible, just creating a whole bunch of submodules. If you have advice on better ways to do this (passing functions outside the module inside so I do not have to redefine it every time, as well as having the using modules propagate), that would be great for future reference.

Super open to suggestions and feedback. This is my first PR in Julia (and ever!) so please be kind. Love the Package and am happy to contribute.

### DIFF
--- a/src/EMNIST/EMNIST.jl
+++ b/src/EMNIST/EMNIST.jl
@@ -1,0 +1,276 @@
+export EMNIST
+module EMNIST
+    using DataDeps
+    using BinDeps
+    using FixedPointNumbers
+    using MAT: matopen, matread
+    using ..MLDatasets: bytes_to_type, datafile, download_dep, download_docstring
+    using ..MNIST: convert2image
+
+        const DEPNAME = "EMNIST"
+
+        """
+            download([dir]; [i_accept_the_terms_of_use])
+
+        Trigger the (interactive) download of the full dataset into
+        "`dir`". If no `dir` is provided the dataset will be
+        downloaded into "~/.julia/datadeps/$DEPNAME".
+
+        This function will display an interactive dialog unless
+        either the keyword parameter `i_accept_the_terms_of_use` or
+        the environment variable `DATADEPS_ALWAY_ACCEPT` is set to
+        `true`. Note that using the data responsibly and respecting
+        copyright/terms-of-use remains your responsibility.
+        """
+        download(args...; kw...) = download_dep(DEPNAME, args...; kw...)
+
+        function __init__()
+            register(DataDep(
+                DEPNAME,
+                """
+                Dataset: The EMNIST Dataset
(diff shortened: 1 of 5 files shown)


src/EMNIST/EMNIST.jl
```diff
@@ -0,0 +1,276 @@
...
+                the website to make sure you want to download the
+                dataset.
+                """,
+                "http://www.itl.nist.gov/iaui/vip/cs_links/EMNIST/matlab.zip",
+                "e1fa805cdeae699a52da0b77c2db17f6feb77eed125f9b45c022e7990444df95",
+                post_fetch_method = file -> (run(BinDeps.unpack_cmd(file,dirname(file),".zip","")); rm(file))
+            ))
+        end
+
+        module balanced
```
johnnychen94, line 61: To avoid confusion, I think we should stick to the conversion that the module name is uppercased. `EMNIST.Balanced`, `EMNIST.ByMerge`, `EMNIST.ByClass`, `EMNIST.Letters`, `EMNIST.Digits`, and `EMNIST.MNIST`. This looks clearer to me, how do you think?

johnnychen94 (commented):
Excellent work! My only concern is the miss of documentation; unlike other datasets (e.g., `MNIST.traindata`, EMNIST provides six types of sub-datasets, e.g., `EMNIST.balanced.traindata` and `EMNIST.digits.traindata`. This could be a source of confusion so I hope there's some usage documentation to it.

src/EMNIST/EMNIST.jl
andrew-saydjari: Agreed. The module names have been updated. I have also added some clarification on the front readme page on how the modules are nested for EMNIST since it is sort of 6x as many datasets as the rest. I have also added a table for fast look up of the number of samples (rather than forcing people to go all the way to the EMNIST webpage for that). Does that solve both problems?

johnnychen94 (approved):

johnnychen94:
@CarloLucibello I noticed that Flux has planed to deprecate its dataset loading codes, so it would be nice to have you double-check this.

johnnychen94:
@andrew-saydjari Thank you for doing this! This will be available soon in MLDatasets v0.5.5

PR #56, opened 2021-02-25
Add FastAI datasets

lorenzoh:
This adds datadeps of all datasets from the [FastAI dataset collection](https://course.fast.ai/datasets).

The basic functionality:

```julia
using MLDatasets.FastAIDatasets
using MLDatasets.FastAIDatasets:
    datasetpath,  # download dataset and get directory
    DATASETS,  # list of all datasets
    loaddataclassification,  # load an image classification dataset into a data container with observations `(image, class)` 
    loaddatasegmentation,  # load an image segmentation dataset into a data container with observations `(image, mask)`  


```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,11 +5,17 @@ version = "0.5.5"
 [deps]
 BinDeps = "9e28174c-4ba2-5203-b857-d8d62c4213ee"
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
+Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
 DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
 DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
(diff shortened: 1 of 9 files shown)


CarloLucibello:
What's the reason for having `loaddatasegmentation` / `loaddataclassification` and not just `loaddata`?

lorenzoh:
They're supposed to work on any folder containing the dataset in the right format, not just the included datasets. Also some datasets can be used for multiple different tasks, so there is no 1-to-1 mapping.

MariusDrulea:
Is this PR still desired/feasible?

lorenzoh:
I will leave the decision on whether this is desirable to more active maintainers of this repository, but will note that implementation-wise a lot of things have changed since this PR was first opened. The largest change is that LearnBase.jl+MLDataPattern.jl have been superseded by MLUtils.jl.
From the FastAI.jl side, I am always happy to take stuff out and move it into a more canonical package, as would be the case with these datasets.

CarloLucibello:
I would be in favor of moving these datasets here, provided we manage to make the interface consistent with the other datasets here. I don't know how hard that would be.

PR #67, opened 2021-07-23
Simplify and update documentation

CarloLucibello:
Simplify the readme, add missing doc pages, generally reduce redundancies


.github/workflows/Documenter.yml
```diff
@@ -14,9 +14,8 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
-      matrix:
-        julia-version: [1]
-        os: [ubuntu-latest]
+      julia-version: 1.6
+      os: ubuntu-latest
```
johnnychen94, line 18: I feel like `1` is a more compatible choice since it automatically uses the latest 1.x julia release.

johnnychen94 (commented):
The idea looks good to me, I notice that a large part of it are whitespace changes, it's okay to do so, but could you revert `src/EMNIST/EMNIST.jl` one since it's 100% unrelated whitespace change.

johnnychen94:
Since it's merged, never mind.

CarloLucibello:
sorry, went too much in a hurry, dealing with this documentation stuff makes me want to be done as soon as possible

.github/workflows/Documenter.yml
CarloLucibello: I'll do it

CarloLucibello:
> The idea looks good to me, I notice that a large part of it are whitespace changes, it's okay to do so, but could you revert src/EMNIST/EMNIST.jl one since it's 100% unrelated whitespace change.

That was reducing the indentation by one level, yeah not related to docs, but there was no code change

PR #71, opened 2021-07-27
update doc CI

CarloLucibello:
hopefully should fix the deployment of stable docs


johnnychen94:
Yep, I think the issue here is that we need to set up ssh key so that our tagbot can trigger the documenter CI.

FYI, I've just added the `DOCUMENTER_KEY` secret.

Issue #74, opened 2021-08-16
All links to datasets documentation in the home page are broken


sylvaticus:
All links in the "datasets" section of the home page (that should point to the specific dataset documentation) are broken (404 not found error), e.g. https://juliaml.github.io/MLDatasets.jl/latest/datasets/MNIST/

johnnychen94:
Mind opening a PR for it?

```diff
- https://juliaml.github.io/MLDatasets.jl/latest/datasets/MNIST/
+ https://juliaml.github.io/MLDatasets.jl/dev/datasets/MNIST/
```

sylvaticus:
Done (as commented in the PR I do believe however that the documentation by default should point to the stable branch)

PR #75, opened 2021-08-16
Update links to datasets documentation in the homepage

sylvaticus:
As suggested in https://github.com/JuliaML/MLDatasets.jl/issues/74

(However, personally I believe it is better to point new users to the "stable" documentation branch, i.e. those that correspond to "add  PkgX")


johnnychen94 (approved):

johnnychen94:
In most times we do continuous releases once there's a new PR in so there shouldn't be much difference between dev and stable branch.

I'm okay with both.

PR #81, opened 2021-10-23
Use Pickle in favor of PyCall

yuehhua:


yuehhua:
Now, it is loosen to v1.3.

johnnychen94:
Thank you all for doing this!

cossio:
Can we tag a release with this? I am having some issues with PyCall which I think would be fixed by this.

johnnychen94:
Done.

cossio:
Thanks!

PR #83, opened 2021-10-31
render black digits on a whiteboard

ShuhuaGao:
Correct the display such that "
the digits are black on a white background".

### DIFF
--- a/src/MNIST/utils.jl
+++ b/src/MNIST/utils.jl
@@ -28,5 +28,5 @@ function convert2image(array::AbstractArray{T}) where {T<:Number}
     else
         img = _colorview(Gray, array)
     end
-    img
+    return one(eltype(img)) .- img
 end
--- a/src/MNIST/utils.jl
+++ b/src/MNIST/utils.jl
@@ -1,8 +1,8 @@
 """
-    convert2image(array) -> Array{Gray}
+    convert2image(array; black_digits=false) -> Array{Gray}
 
 Convert the given MNIST horizontal-major tensor (or feature matrix)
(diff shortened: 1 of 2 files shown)


CarloLucibello:
Sorry we can't accept this PR, it's a breaking change not worth doing

Evizero:
interesting. I think to remember that this actually used to be implemented that way. If this changed for some reason, which seems to be the case, then the docs should probably reflect it. They still claim:

```
The values are also color
corrected according to the website's description, which means that
the digits are black on a white background.
```

ShuhuaGao:
Yes, I have the same point as @Evizero. The current implementation does not match the doc: either the code or the doc should be updated.

CarloLucibello:
Sorry, didn't notice that. Change the docs?

ShuhuaGao:
An alternative way is to introduce a keyword argument: `black_digits=false`. It will not cause breaking changes.

CarloLucibello:
yes, that would be fine. The docs should reflect the default behavior though

ShuhuaGao:
@CarloLucibello Hi, any advice on this PR? The default behaviour remains unchanged to avoid breaking changes, but the user has the freedom to render black digits on a white background.

CarloLucibello:
Looks fine, can you add a test? e.g. that the returned array is reversed compared to the original one?

ShuhuaGao:
@CarloLucibello Done. No idea why "Unit test / test (1.0, windows-latest) (pull_request) " failed. It does not seem to be caused by my changes.

CarloLucibello:
thanks @ShuhuaGao

PR #84, opened 2021-11-05
removing labels from features

racinmat:
This commits removes labels from features, because previously the labels were part of the features, which is fundamentally wrong. 
Data is materialized as a side-effect, because JSON3 does not provide a way to lazily delete this, but it should not be problem since the dataset is very small.

### DIFF
--- a/src/Mutagenesis/Mutagenesis.jl
+++ b/src/Mutagenesis/Mutagenesis.jl
@@ -22,13 +22,12 @@ julia> test_x, test_y = Mutagenesis.testdata();
 julia> val_x, val_y = Mutagenesis.valdata();
 
 julia> train_x[1]
-JSON3.Object{Base.CodeUnits{UInt8, String}, SubArray{UInt64, 1, Vector{UInt64}, Tuple{UnitRange{Int64}}, true}} with 6 entries:
+Dict{Symbol, Any} with 5 entries:
   :ind1      => 1
   :inda      => 0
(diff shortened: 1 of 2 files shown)


CarloLucibello:
Ok, so this can be considered a bug fix and the next tagged release non-breaking

CarloLucibello:
Also materializing makes everything more convenient. Is everything materialized now?

racinmat:
Yes, it's a bug fix.
No, it's not materialized now and it's a bit annoying to deal with. That's why I wanted to use JSON instead of JSON3 originally.

CarloLucibello:
Can we just have the interface return only arrays or dictionaries?

racinmat:
Now, when the `copy` is present, it returns only arrays and dictionaries, no JSON3 structures are present in the returned data. 
Or did you have something else on mind?

CarloLucibello:
just that. Can you add type checks to tests like `@test train_x isa AbstractArray` or similars?

racinmat:
Yes, I'll add it, good idea.

racinmat:
Added the tests are now stricter, because `isa AbstractArray` would have passed even without the materialization, with JSON3 lazy loading.

CarloLucibello:
The docstring needs to be changed (you can see that from the Documentation CI failure)

racinmat:
Fixed.

CarloLucibello:
thanks!

Issue #73, opened 2021-07-28
redesign the package


CarloLucibello:
I'd like to open a discussion on how we should move forward with implementing a `getobs` and `nobs` compliant api, 
while possibly also simplifying the interface and the maintenance burden.  

See  https://github.com/FluxML/FastAI.jl/issues/22

I think we should move away from the module-based approach and adopt a type-based one. Also could be convenient to have some lean type hierarchy. 

Below is an initial proposal

## AbstractDatasets
```julia
####### src/datasets.jl

abstract type AbstractDataset end
abstract type FileDataset <: AbstractDataset end
abstract type InMemoryDataset <: AbstractDataset end
```
## MNIST Dataset

```julia
######  src/vision/mnist.jl
"""
docstring here, also exposing the internal fields of the struct for transparency
"""
struct MNIST <: InMemoryDataset
   x                              # alternative names: `features` or `inputs`
   targets                     #                               `labels` or j`y`
   num_classes           # optional
   
    function MNIST(path=nothing; split = :train)  # split could be made a mandatory keywork arg
          @assert split in [:train, :test]
           ..........
   end
end

LearnBase.getobs(data::MNIST) = (data.x, data.target) 
LearnBase.getobs(data::MNIST, idx) = (data.x[:,idx], data.target[idx]) 
LearnBase.nobs(data::MNIST) = length(data.taget)

.... other stuff ....
```
## Usage
```julia
using MLDasets: MNIST
using Flux.
train_data = MNIST(split = :train)
test_data = MNIST(split =:test)

xtrain, ytrain = getobs(train_data)
xtrain, ytrain = train_data # we can add this for convenience

xs, ys = getobs(train_data, 1:10)
xs, ys = train_data[1:10] # we can add this for convenience


train_loader = DataLoader(train_data; batch_size=128)
``` 
## Transforms

Do we need transformations as part of the datasets?
This is a possible interface that assumes the transform to operate on whatever is returned by `getobs`

```julia
getobs(data::MNIST, idx) = data.transform(data.x[:,idx], data.y[idx])

Data(split = :train, transform = (x, y) -> (random_crop(x), y)  
```

## Deprecation Path 1
We can create a deprecation path for the code
```julia
using MLDataset: MNIST
xtrain, ytrain = MNIST.traindata(...)
```
by implementing
```
function getproperty(data::MNIST, s::Symbol)
  if s == :traindata
    @warn "deprecated method"
    return ....

  ....
end
``` 

[...]

PR #96, opened 2022-02-19
Redesign package to be built on top of reusable dataset containers

darsnack:
This works towards #73 by adding low-level dataset containers that are MLUtils.jl compatible out of the box. The end goal is to rewrite the existing datasets to use these containers, but this PR will just add the containers without changing the existing functionality that users expect. In a followup PR, I will rewrite the existing API, provide a deprecation path, and add better documentation (docstrings will still be added here first).

So far, I have ported only the containers from FastAI.jl. I intend to implement the following before merging this PR:

**Low level**
- [x] `FileDataset`: lazy access to a vector of files on disk
- [x] `TableDataset`: any Tables.jl compatible table
- [x] Text: this one I need to decide how to best reuse JuliaText similar to how `TableDataset` wraps Tables.jl
- [x] `HDF5Dataset`: built on HDF5.jl
- [x] `JLD2Dataset`: built on JLD2.jl
- [x] ~~`MATDataset`: built on MAT.jl~~ the data can just be read in as a `Dict` and used with MLUtils.jl without any extra effort (there is no value to having a wrapper here)

**Mid level**
- [x] `CachedDataset`: wraps a lazy dataset like `FileDataset` so that it stays in memory (this is how small datasets like MNIST will be)
- [x] ~~`GroupedDataset`: wraps several named containers together (e.g. for "train" + "val" data)~~ MLUtils already has functionality for this


CarloLucibello:
Given https://github.com/JuliaML/MLUtils.jl/pull/56, should we avoid depending on MLUtils and just define `length` and `getindex`?

CarloLucibello:
We should drop julia < 1.6

darsnack:
The way I wrote `AbstractDataContainer` makes the reverse true as well: anything that supports `getobs`/`numobs` automatically gets `getindex`, `length`, and `iterate`.

If the MLUtils.jl dependency is too expensive, I'd rather move ahead with splitting the interface into its own base package. My view is that MLDatasets should be a first-class consumer of the MLUtils interface. I also think that MLUtils might be a cheaper dep than some of the other IO related packages. Though this is a good question to consider: should there be a "just the data deps" package that splits out only the download/unpack functions?

darsnack:
I decided against incorporating text in this PR since it involves a temporal dimension that I haven't yet discussed as part of the MLUtils.jl refactor. I also expect that the best option will involve factoring out text datasets similar to how we discussed for vision.

Would also be useful to get @lorenzoh's eyes on this PR.

src/containers/cacheddataset.jl
```diff
@@ -0,0 +1,39 @@
...
+By default the observation indices `1:cachesize` are cached.
+You can manually pass in a `cache` and set of `cacheidx` as well.
+
+See also [`make_cache`](@ref) for customizing the default cache creation for `source`.
+"""
+struct CachedDataset{T, S}
+    source::T
+    cacheidx::Vector{Int}
+    cache::S
+end
```
CarloLucibello, line 26: Should this go in MLUtils.jl?

src/containers/cacheddataset.jl
```diff
@@ -0,0 +1,39 @@
...
+
+    CachedDataset(source, collect(cacheidx), make_cache(source, cacheidx))
+end
+
+function MLUtils.getobs(dataset::CachedDataset, i::Integer)
+    _i = findfirst(==(i), dataset.cacheidx)
+
+    return isnothing(_i) ? getobs(dataset.source, i) : getobs(dataset.cache, _i)
+end
+MLUtils.numobs(dataset::CachedDataset) = numobs(dataset.source)
```
CarloLucibello, line 39:
```suggestion
Base.length(dataset::CachedDataset) = numobs(dataset.source)
```

src/containers/cacheddataset.jl
```diff
@@ -0,0 +1,39 @@
...
+    cache::S
+end
+
+function CachedDataset(source, cachesize::Int = numobs(source))
+    cacheidx = 1:cachesize
+
+    CachedDataset(source, collect(cacheidx), make_cache(source, cacheidx))
+end
+
+function MLUtils.getobs(dataset::CachedDataset, i::Integer)
```
CarloLucibello, line 34:
```suggestion
function Base.getindex(dataset::CachedDataset, i::Integer)
```

src/containers/cacheddataset.jl
```diff
@@ -0,0 +1,39 @@
...
+"""
+    CachedDataset(source, cachesize = numbobs(source))
+    CachedDataset(source, cacheidx, cache)
+
+Wrap a `source` data container and cache `cachesize` samples in memory.
+This can be useful for improving read speeds when `source` is a lazy data container,
+but your system memory is large enough to store a sizeable chunk of it.
+
+By default the observation indices `1:cachesize` are cached.
+You can manually pass in a `cache` and set of `cacheidx` as well.
```
CarloLucibello, line 18: Not clear what `cache` is supposed to be here.

src/containers/cacheddataset.jl
```diff
@@ -0,0 +1,39 @@
...
+See also [`make_cache`](@ref) for customizing the default cache creation for `source`.
+"""
+struct CachedDataset{T, S}
+    source::T
+    cacheidx::Vector{Int}
+    cache::S
+end
+
+function CachedDataset(source, cachesize::Int = numobs(source))
+    cacheidx = 1:cachesize
```
CarloLucibello, line 29: storing the first chachesize entries seems totally arbitrary.

src/containers/filedataset.jl
```diff
@@ -0,0 +1,57 @@
...
+FileDataset(loadfn,
+            dir::Union{AbstractPath, AbstractString},
+            pattern::AbstractString = "*",
+            depth = 4) = FileDataset(loadfn, rglob(pattern, string(dir), depth))
+FileDataset(dir::Union{AbstractPath, AbstractString}, pattern::AbstractString = "*", depth = 4) =
+    FileDataset(loadfile, dir, pattern, depth)
+
+MLUtils.getobs(dataset::FileDataset, i::Integer) = loadfile(dataset.paths[i])
+MLUtils.getobs(dataset::FileDataset, is::AbstractVector) = map(Base.Fix1(getobs, dataset), is)
+MLUtils.numobs(dataset::FileDataset) = length(dataset.paths)
```
CarloLucibello, line 57: `getindex` and `length`

src/containers/filedataset.jl
```diff
@@ -0,0 +1,57 @@
...
+const RE_IMAGEFILE = r".*\.(gif|jpe?g|tiff?|png|webp|bmp)$"i
+isimagefile(f) = matches(RE_IMAGEFILE, f)
+
+"""
+    rglob(filepattern, dir = pwd(), depth = 4)
+
+Recursive glob up to `depth` layers deep within `dir`.
+"""
+function rglob(filepattern = "*", dir = pwd(), depth = 4)
+    patterns = [repeat("*/", i) * filepattern for i in 0:(depth - 1)]
```
CarloLucibello, line 12: does this work on all platforms?

src/containers/filedataset.jl
```diff
@@ -0,0 +1,57 @@
...
+
+"""
+    loadfile(file)
+
+Load a file from disk into the appropriate format.
+"""
+function loadfile(file::String)
+    if isimagefile(file)
+        # faster image loading
+        return FileIO.load(file, view = true)
```
lorenzoh, line 25: I think this special case is no longer necessary with ImageIO.jl now defaulting to JpegTurbo.jl (`view` is ImageMagick specific I think.

src/containers/filedataset.jl
```diff
@@ -0,0 +1,57 @@
...
+
+    return vcat([glob(pattern, dir) for pattern in patterns]...)
+end
+
+"""
+    loadfile(file)
+
+Load a file from disk into the appropriate format.
+"""
+function loadfile(file::String)
```
lorenzoh, line 22: Not sure if a bunch of nested `if`s is the most elegant solution for this. Since this was meant to be a high-level function for FastAI.jl, maybe we can do without it for now?

lorenzoh (commented):
Left some comments, mostly on the uglier bits I put into FastAI.jl.

Beside those things, I think we should standardize on the type we use to represent file paths, i.e. `String` or `AbstractPath`. Even if we want to allow passing both, I think it makes sense to use the  same repr. internally.

In favor of `String`s:

- bit simpler
- the globbing only returns files and I think converting these to `Path`s consumes more memory

In favor of `AbstractPath`s:
- more thoughtful API
- more extensible (e.g. other types of file systems like S3)
- less ambiguous

What do you think?

src/containers/cacheddataset.jl
darsnack: Any subset of the indices would be arbitrary. The main reason for including this is to address FFCV from [this issue](https://github.com/JuliaML/MLUtils.jl/issues/13). My lab mate is using FFCV with PyTorch on ImageNet right now. We strongly suspect that the overwhelming majority of the performance gains are from caching a portion of the dataset in memory (my lab mate will request 400GB on our cluster). In this case, the particular indices don't matter so much and how many, so the first N seems as good as any.

src/containers/cacheddataset.jl
darsnack: I can make a PR though this is only useful for data that isn't already in memory. I had trouble thinking of cases where that's true but the data isn't a dataset.

src/containers/filedataset.jl
darsnack: Should I just go with `FileIO.load` then?

darsnack:
I went with `AbstractString` because it doesn't seem like FilePathsBase.jl is used frequently enough in the ecosystem.

src/containers/filedataset.jl
lorenzoh: Yeah, I think that's the best for now. But let's re-export `load` instead of creating a new `loadfile` function

ablaom:
@darsnack Do the wrapped `TableDataset` objects themselves implement the Tables.jl interface? That is, is every `TableDataset(some_table)` also a table?

darsnack:
Not at this moment, but based on our discussion it's clear that it should! I'll add this tomorrow.

ablaom:
If not already, I suggest the wrapper be parameterised by the type of the table being wrapped.  See continuation of same discussion.

darsnack:
Yeah for exactly the reasons in the discussion, we already do this. And we have faster paths for when the underlying table is a `DataFrame` or `CSV.File`.

ablaom:
I see. That works here, but not ideal if we were to  move this to MLUtils (say)  because we would need DataFrames and CSV as dependencies. Those `getobs` implementations should ideally live in those packages (as does the Tables interface), right?

src/containers/tabledataset.jl
```diff
@@ -0,0 +1,58 @@
...
+    function TableDataset{T}(table::T) where {T}
+        Tables.istable(table) ||
+            throw(ArgumentError("TableDatasets must implement the Tabels.jl interface"))
+
+        new{T}(table)
+    end
+end
+
+TableDataset(table::T) where {T} = TableDataset{T}(table)
+TableDataset(path::AbstractPath) = TableDataset(DataFrame(CSV.File(path)))
```
lorenzoh, line 22:
```suggestion
TableDataset(path::AbstractString) = TableDataset(DataFrame(CSV.File(path)))
```

lorenzoh: I think this is what is keeping the tests from passing.

CarloLucibello:
CI fails due to a missing `using FileIO` I guess. Besides that, this looks good to me

darsnack:
Okay this is probably good to go now. `TableDataset` is now a Tables.jl table. We can continue the discussion around tables in MLUtils.jl and update here as needed. But for now, this will allow us to move forward in this repo with the refactor.

CarloLucibello:
You can remove  `strict = true` from docs/make.jl if you don't want to add this stuff to the documentation right now.

darsnack:
I just added a WIP page so that the docstrings are at least there should someone choose to use the dev branch. I also changed the behavior to only check docstrings for exported functions (since I added `rglob` which we probably don't want in the actual docs).

CarloLucibello (approved):

PR #98, opened 2022-03-20
add SupervisedDataset and move misc. datasets to the new interface

CarloLucibello:
This is a follow-up to #96 and a continuation of the redesign planned in #73, aiming at redesigning only the simple misc. datasets (Titanic, Iris, BostonHousing, and Mutagenesis) for the time being.

~`ArrayDataset` is introduced for completeness but not used, I think it is better to have flat structures whenever we can.~

~`AbstractDataset`~ `SupervisedDataset` is introduced to remove some boilerplate code from each dataset definition and enforce a consistent interface. It gives:
- An informative print of a new dataset with zero effort.
- Indexing behavior assuming the dataset contains the fields `features` and `targets`. 

Indexing or getobs produce a ~named~ tuple `(features, targets) `, ~which seems to be more informative than just returning a tuple and still allows `x, y = dataset[i]`.~

### Deprecations
The deprecation allows the following old code to work:
```julia
using MLDataset: BostonHousing

X = BostonHousing.features()
```
but the following will be broken:
```julia
using MLDataset.BostonHousing: features

X = features()
```
Each old test is preserved and passes.


src/MLDatasets.jl
```diff
@@ -17,34 +17,14 @@ import MLUtils
...
-    ImageCore.channelview(image)
-end
-
-function _colorview(::Type{T}, array::AbstractArray{<:Number}) where T <: Color
-    __images_supported__ ||
-        error("Converting to image requires `using ImageCore`")
-    ImageCore.colorview(T, array)
-end
-
-
```
CarloLucibello, line 45: these are just moved into the newly created `utils.jl`

src/containers/cacheddataset.jl
```diff
@@ -20,15 +20,17 @@ You can manually pass in a set of `cacheidx` as well.
...
     source::T
     cacheidx::Vector{Int}
     cache::S
 end
 
+CachedDataset(source, cachesize::Int) = CachedDataset(source, 1:cachesize)
+
 CachedDataset(source, cacheidx::AbstractVector{<:Integer} = 1:numobs(source)) =
     CachedDataset(source, collect(cacheidx), make_cache(source, cacheidx))
-CachedDataset(source, cachesize::Int = numobs(source)) = CachedDataset(source, 1:cachesize)
```
CarloLucibello, line 31: this is definition seems ambiguous with the one above

src/containers/cacheddataset.jl
```diff
@@ -20,15 +20,17 @@ You can manually pass in a set of `cacheidx` as well.
 
 See also [`make_cache`](@ref) for customizing the default cache creation for `source`.
 """
-struct CachedDataset{T, S}
+struct CachedDataset{T, S} <: AbstractDataContainer
```
CarloLucibello, line 23: I think this was just forgotten, every other container has it

src/datasets_misc/boston_housing.jl
```diff
@@ -41,77 +43,76 @@ Censoring is suggested by the fact that the highest median price of exactly \\\$
...
+#     elseif s == :targets
+#         Base.depwarn("BostonHousing.targets() is deprecated, please use `BostonHousing().targets` instead", :targets)
+#         return d.targets
+#     elseif s == :feature_names
+#         Base.depwarn("BostonHousing.targets() is deprecated, please use `BostonHousing().feature_names` instead", :feature_names)
+#         return d.feature_names
+#     else 
+#         return getfield(d, s)
+#     end
+# end
```
CarloLucibello, line 118: uncommenting in this, I get an error already at object creation, julia tries to access T.name and is rewired here.
```julia
julia> dataset = BostonHousing()
s = :name
Error showing value of type s = :name
ERROR: type BostonHousing has no field name
Stacktrace:
  [1] getproperty(#unused#::s = :name

SYSTEM (REPL): showing an error caused an error
ERROR: type BostonHousing has no field name
Stacktrace:
  [1] getproperty(#unused#::s = :name

SYSTEM (REPL): caught exception of type ErrorException while trying to handle a nested exception; giving up
```

CarloLucibello: nevermind, there was a silly mistake

darsnack (changes requested):
I agree that `ArrayDataset` is not necessary and it is better to just use the array directly.

In general, I think we should try to minimize directly reading and writing the files on disk as much as possible. If the low level containers don't automate some of this for us, then they probably won't be useful to users either. We want something like:
```julia
function Iris(; dir = nothing)
  file = datafile("Iris", "iris.data", dir)
  # table should already be cached in memory via DataFrame
  # accessing rows already produces a named tuple of the columns
  # could have TableDataset accept a "colnames" keyword if we want
  # to always have features/targets
  data = TableDataset(file)

  Iris(file, data)
end
```

Maybe we can have it so that
```julia
struct Iris <: AbstractDataset
  # ...
end

# datasets = container + metadata
# interface: get the core container
samples(iris::Iris) = iris.data

# boilerplate
getindex(dataset::AbstractDataset, i) = getobs(samples(dataset), i)
length(dataset::AbstractDataset) = numobs(samples(dataset))
```

darsnack:
The goal here is to have a unified interface for in-memory table datasets, right? I'm slightly confused what the interface is as per the PR. Perhaps it makes more sense to not have `AbstractDataset` and instead have `SupervisedDataset` which requires the functions `features` and `targets` defined as well as the named tuple `getobs` discussed above?

Sidenote: the fact that Boston Housing or Titanic require a `feature_names` field seems like a strong indicator to me that they should be tables like a `DataFrame` and not auto-convert to arrays.

darsnack (changes requested):
Looks good to me, but I would suggest dropping the `as_df` keyword for now (reverting to always returning arrays). Maybe we can consider `DataFrame` support a feature for later, so that we don't introduce something into the API now that we need to potentially deprecate.

src/datasets_misc/mutagenesis.jl
```diff
@@ -0,0 +1,122 @@
...
+    DEPNAME = "Mutagenesis"
+    DATA = "data.json"
+    METADATA = "meta.json"
+
+    @assert split ∈ [:train, :val, :test, :all]
+    
+    data_path = datafile(DEPNAME, DATA, dir)
+    metadata_path = datafile(DEPNAME, METADATA, dir)
+    samples = open(JSON3.read, data_path)
+    metadata = open(JSON3.read, metadata_path)
```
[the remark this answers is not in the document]
CarloLucibello: I'm not using those containers yet

src/abstract_dataset.jl
```diff
@@ -0,0 +1,19 @@
...
+end
+
+_summary(x) = x
+_summary(x::AbstractArray) = summary(x)
+
+
+# TODO check if hasfield(D, :targets) to handle both supervised and unsupervised
+Base.getindex(d::AbstractDataset) = getobs((; d.features, d.targets))
+Base.getindex(d::AbstractDataset, i) = getobs((; d.features, d.targets), i)
+Base.length(d::AbstractDataset) = numobs(d.features)
```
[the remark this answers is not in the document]
CarloLucibello: done

CarloLucibello:
I'm happy with this actually. Unless dealing with very very large tables, we generically want to load tables as dataframes, no point in having lazy wrappers. Also, array representations are not ok in most cases, you lose per-column eltype information and also have to store column names elsewhere. For data wrangling of tabular data, you just want data frames. 
So here I'm just mimicking the `sklearn.dataset` interface and we add on top of that the indexing/getobs interface. 
I think we just want the `as_df` option, and hardly go back, so since it is ready let's just do it.

CarloLucibello:
Unfortunately, first-time use of `CSV.read` has a lot of compilation overhead:

 ```julia
julia> @time CSV.read("../../datadeps/Iris/iris.data", DataFrame, header=0)
 14.600421 seconds (2.93 M allocations: 129.822 MiB, 99.98% compilation time)
```
So this PR gives
```julia
# MASTER
julia> @time using MLDatasets
  6.931466 seconds (14.01 M allocations: 824.053 MiB, 6.86% gc time, 69.56% compilation time)

julia> @time Iris().features;
 15.290096 seconds (4.65 M allocations: 226.622 MiB, 0.25% gc time, 99.98% compilation time)

julia> @time Iris().features;
  0.000516 seconds (512 allocations: 42.750 KiB)
```
While on master, where `readdlm` is used instead, we have 
```julia
julia> @time using MLDatasets
  6.996220 seconds (14.53 M allocations: 839.009 MiB, 6.59% gc time, 69.33% compilation time)
  
julia> @time Iris.features();
  0.874926 seconds (1.71 M allocations: 95.881 MiB, 5.78% gc time, 99.82% compilation time)

julia> @time Iris.features();
  0.001535 seconds (15.43 k allocations: 461.656 KiB)

CarloLucibello:
I'm going on and merging this since there are a lot of datasets still to be ported and plenty of time for more discussions and revisiting choices before the next tag. @darsnack if you are unhappy with this design let's continue the discussion in #73 or even better let's open an issue specific to in-memory tabular datasets

darsnack:
Sounds good, sorry I missed your messages from 2 days ago for some reason. I'll open an issue for further discussion.

PR #112, opened 2022-04-13
Add Zachary's Karate Club Dataset

Dsantra92:
Added [Zachary's Karate Club Dataset](https://en.wikipedia.org/wiki/Zachary%27s_karate_club). The data was taken from PyG [Karate Club implementation](https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/datasets/karate.html#KarateClub). There mainly two different variations for the Karate Club:
1. The dataset's original karate club labels. There are 3 unique labels. The node label depends on whether a student stayed with Hi or not. Or whether they chose to leave karate altogether. [NetworkX](https://networkx.org/documentation/stable/auto_examples/graph/plot_karate_club.html) and some data providers that I have come across provide this variation.
2. The clustering-based labels. In this case, there are 4 unique labels. These labels are obtained from the graph using Modularity-based clustering. PyG uses this variation.

This implementation follows the 2nd variation.


CarloLucibello:
nice! Can you shift labels and edge_index by 1 so that we are consistent with julia 1-based indexing?

Does the edge_index contain the edges in both directions, i.e. twice the number of undirected edges? it should

Can we also provide the original labeling, maybe by adding a keyword arg to the `labels` function?

Dsantra92:
Adding a keyword for the labels is nice idea. Yes the number of edges in the adjacency list is 2 x number of edges. The 1-based indexing is nice catch, I am updating the node labels and edge indexes accordingly.

CarloLucibello:
Tests are missing

test/runtests.jl
```diff
@@ -33,6 +33,7 @@ dataset_tests = [
     "datasets_graph/deprecated_pubmed.jl",
     "datasets_graph/deprecated_tudataset.jl",
     "datasets_graph/deprecated_polblogs.jl",
+    "datasets_graph/karteclub.jl",
```
CarloLucibello, line 36:
```suggestion
    "datasets_graph/karateclub.jl",
```

CarloLucibello:
Thanks!

PR #128, opened 2022-05-10
lazy module loading

### LINKED ISSUE #126: using MLDatasets is very slow
In a fresh julia 1.7 session
```julia
julia> @time using MLDatasets
 13.485246 seconds (20.31 M allocations: 1.158 GiB, 7.56% gc time, 61.69% compilation time)
```
Is there a way to conditionally import packages?
```julia
julia> for pkg in [:ImageCore, :CSV, :HDF5, :JLD2, :JSON3]; print(pkg); @time @eval using $pkg; end
ImageCore  2.141235 seconds (3.02 M allocations: 200.377 MiB, 4.50% gc time, 32.08% compilation time)
CSV  3.817959 seconds (6.14 M allocations: 348.493 MiB, 9.81% gc time, 90.11% compilation time)
HDF5  0.723358 seconds (1.34 M allocations: 73.225 MiB, 1.69% gc time, 93.80% compilation time)
JLD2  1.139716 seconds (1.36 M allocations: 78.966 MiB, 3.95% gc time, 60.77% compilation time)
JSON3  0.033367 seconds (49.09 k allocations: 3.014 MiB)

julia> for pkg in [:DataFrames, :MLUtils, :Pickle, :NPZ, :MAT]; print(pkg); @time @eval using $pkg; end
DataFrames  1.789793 seconds (2.03 M allocations: 137.197 MiB, 4.63% gc time)
MLUtils  1.743072 seconds (2.07 M allocations: 117.900 MiB, 4.83% gc time, 47.32% compilation time)
Pickle  0.130685 seconds (159.17 k allocations: 9.751 MiB, 17.77% compilation time)
NPZ  0.504406 seconds (1.19 M allocations: 61.838 MiB, 4.05% gc time, 98.87% compilation time)
MAT  0.009792 seconds (22.84 k allocations: 1.044 MiB)
```

Related [discourse thread](https://discourse.julialang.org/t/import-a-module-only-when-needed/80654)

CarloLucibello:
Fix #126 using FileIO and a similar method for lazy loading modules. 
Time-to-first-MNIST goes from 14.5s to 5.5s.
 
```julia
###### MASTER ##############

julia> @time using MLDatasets
 12.421854 seconds (19.82 M allocations: 1.144 GiB, 6.86% gc time, 61.53% compilation time)

julia> @time MNIST()
  1.965113 seconds (1.34 M allocations: 301.964 MiB, 13.25% gc time, 44.99% compilation time)
dataset MNIST:
  metadata    =>    Dict{String, Any} with 3 entries
  split       =>    :train
  features    =>    28×28×60000 Array{Float32, 3}
  targets     =>    60000-element Vector{Int64}

#### THIS PR #####################
julia> @time using MLDatasets
  3.757601 seconds (5.07 M allocations: 293.185 MiB, 5.89% gc time, 66.09% compilation time)

julia> @time MNIST()
  1.653665 seconds (1.33 M allocations: 299.937 MiB, 3.27% gc time, 48.02% compilation time)
dataset MNIST:
  metadata    =>    Dict{String, Any} with 3 entries
  split       =>    :train
  features    =>    28×28×60000 Array{Float32, 3}
  targets     =>    60000-element Vector{Int64}

```


CarloLucibello:
@johnnychen94 I'm running in some world age issue, see the failing tests. Do you know how to fix them?

src/io.jl
```diff
@@ -9,18 +11,31 @@ function read_csv(path, sink::Type{A}; kws...) where A <: AbstractMatrix
...
+    return FileIO.load(File{format"NPZ"}(path))
 end
 
 function read_pytorch(path)
-    return Pickle.Torch.THload(path)
+    Base.invokelatest(checked_import(idPickle).mod.Torch.THload, path)
+end
+
+function read_pickle(path)
+    checked_import(idPickle).npyload(path)
```
johnnychen94, line 32: The world age issue is because many method definitions are not yet "refreshed" when you do dynamic lazy call from an old world.

This is exactly where `invokelatest` is used: it calls the method from the latest world where the methods are refreshed.

`invokelatest` has some small overhead, so if the dependencies are not that big, we can preload them via, e.g., `using Pickle` as we used to do.
But yes, loading a predefined __small__ dataset is definitely a perfect case where `invokelatest` overhead is acceptable.

```suggestion
    Base.invokelatest(checked_import(idPickle).npyload, path)
```

Similar changes should be made elsewhere.

src/imports.jl
```diff
@@ -0,0 +1,39 @@
...
+function Base.getproperty(m::ImportedModule, s::Symbol)
+    if s == :mod
+        return getfield(m, s)
+    else
+        function f(args...; kws...)
+            Base.invokelatest(getproperty(m.mod, s), args...; kws...) 
+        end
+        return f
+    end
+end
```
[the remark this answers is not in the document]
johnnychen94: https://github.com/johnnychen94/LazyModules.jl looks quite interesting to me.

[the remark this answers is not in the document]
johnnychen94: registered https://github.com/JuliaRegistries/General/pull/60058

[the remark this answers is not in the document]
johnnychen94: Yes, to avoid the world-age issue, I see two options:

- load some core packages (e.g., ImageCore) that have the necessary methods implemented to really start the work
- use `invokelatest` everywhere until it gets a non-alien type. By "alien type", I mean the type from the lazily loaded packages. This is quite similar to type instability where you can make it stable by inserting a force type conversion/declaration at some point.

src/imports.jl
```diff
@@ -0,0 +1,41 @@
+# Check that packages like JSON3 are already imported by the user.
+
+function require_import(m::Symbol)
+    pkgid = Base.identify_package(string(m))
```
johnnychen94, line 4: Heads up, I believe I made a big mistake here:

Using `identify_package` might break precompilation: https://github.com/JuliaLang/julia/pull/45287/files#r873871094

johnnychen94: Exactly, as least that's how ImageIO did in the old versions.

src/MLDatasets.jl
```diff
@@ -2,28 +2,35 @@ module MLDatasets
...
 
+# Use `@lazy import SomePkg` whenever the returned types are not its own types,
+# since for methods applied on the returned types we would encounter in world-age issues
+# (see discussion in  https://github.com/JuliaML/MLDatasets.jl/pull/128).
+# In the other case instead, use `require import SomePkg` to force 
+# the use to manually import.
 
+@require import JSON3="0f8b85d8-7281-11e9-16c2-39a750bddbf1"
+@require import DataFrames="a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
+@lazy import ImageCore="a09fc81d-aa75-5fe9-8630-4744c3626534"
```
johnnychen94, line 25: FWIW, I believe we need to eagerly import `Colors` and `ColorVectorSpace` to properly avoid the world-age issue brought by lazily importing `ImageCore`.

src/MLDatasets.jl
```diff
@@ -2,28 +2,35 @@ module MLDatasets
...
+include("require.jl") # export @require
 
+# Use `@lazy import SomePkg` whenever the returned types are not its own types,
+# since for methods applied on the returned types we would encounter in world-age issues
+# (see discussion in  https://github.com/JuliaML/MLDatasets.jl/pull/128).
+# In the other case instead, use `require import SomePkg` to force 
+# the use to manually import.
 
+@require import JSON3="0f8b85d8-7281-11e9-16c2-39a750bddbf1"
+@require import DataFrames="a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
```
johnnychen94, line 24: It seems that this package is loaded by some dependencies already:

```julia
julia> using MLDatasets

julia> dataset = Titanic()
dataset Titanic:
  metadata    =>    Dict{String, Any} with 5 entries
  features    =>    891×11 DataFrame
  targets     =>    891×1 DataFrame
  dataframe   =>    891×12 DataFrame
```

I would expect this to error for the first call but it isn't.

CarloLucibello: for me it is erroring. Are you in the right branch?

johnnychen94: Oh my mistake; I forgot to activate the current folder 🙈

src/MLDatasets.jl
CarloLucibello: Ok, so we can just make ImageCore a `@require`. Or  even better, we could `@require import ImageShow`, and from there we get `ImageShow.ImageCore` and `ImageShow.ImageCore.RGB` and nice display in editors.

johnnychen94 (commented):
The overall change looks good to me although I'm not confident that users want the `@require` error -- I understand how you want to reduce the package loading latency, just unsure if it's a good direction.

Said that, if we introduce the "hard" requirement on package loading, then we might need to add some explaination before/after the https://juliaml.github.io/MLDatasets.jl/stable/#Basic-Usage to tell users how to properly handle this.

Because the hard requirement will break people's code, this will need a 0.7.0 version bump.

src/MLDatasets.jl
johnnychen94: Or even the entire Images https://github.com/JuliaImages/Images.jl/issues/1009, if you'd say so.

But again I'm not very confident that `@require` is a good direction, let's stick to `ImageCore` only for this PR.

CarloLucibello: but ImageShow doesn't add much much on top of ImageCore and it is something that most likely people will import in any case. Having an error message that helps discovering ImageShow is useful (it would have been useful to me https://github.com/fonsp/Pluto.jl/discussions/2097)

johnnychen94: > but ImageShow doesn't add much much on top of ImageCore and it is something that most likely people will import in any case.

This entirely depends on whether you're a user or a developer. If you're a user, we try to make `using Images` the simple answer (and #1009 to lazify the entire Images). If you're a developer, there's no need to depend on ImageShow -- no matter how small it is.

PR #136, opened 2022-05-26
eagerly load CSV

### LINKED ISSUE #135: Possible World Age issue when require error is not invoked
## Introduction

Loading a required package for a dataset can be done in two ways. 

1. Try to load a package $ \rightarrow $ figure out the additional packages from error $ \rightarrow $ import the packages and load it again.
2. Load the required packages before trying to load a dataset.

While the 1st method gives no (final) error, the 2nd method gives possible world age issues. 

## Reproduce

Background: Loading OGB-Datasets requires the `DataFrame` package for processing the downloaded CSVs which contain the metadata for all available datasets. `DataFrames` is a `require` package.

1. Importing DataFrames after I get a require import error (No Error)
 ```julia
using MLDatasets: OGBDataset
data = OGBDataset("ogbg-molhiv") # throws a error for import of DataFrames

using DataFrames # Import correction
data = OGBDataset("ogbg-molhiv") # Works fine
```
2.  Importing DataFrames with MLDatasets and then loading a dataset (Throws Error)
```julia
using MLDatasets, DataFrames
data = OGBDataset("ogbg-molhiv")
```
Error stacktrace:
```julia-repl
ERROR: MethodError: no method matching iterate(::InlineStrings.String31)
The applicable method may be too new: running in world age 31389, while current world is 31398.
Closest candidates are:
  iterate(::InlineStrings.InlineString) at ~/.julia/packages/InlineStrings/F5Dhz/src/InlineStrings.jl:520 (method too new to be called from this world context.)
  iterate(::InlineStrings.InlineString, ::Int64) at ~/.julia/packages/InlineStrings/F5Dhz/src/InlineStrings.jl:520 (method too new to be called from this world context.)
  iterate(::AbstractString, ::Integer) at ~/julia/share/julia/base/strings/basic.jl:157
  ...
Stacktrace:
  [1] print(io::IOBuffer, s::InlineStrings.String31)
    @ Base ./strings/io.jl:191
  [2] print_to_string(xs::InlineStrings.String31)
    @ Base ./strings/io.jl:144
  [3] String(s::InlineStrings.String31)
    @ Base ./strings/string.jl:82
  [4] (::MLDatasets.var"#385#386")(r::DataFrameRow{DataFrame, DataFrames.Index})
    @ MLDatasets ./none:0
  [5] iterate
    @ ./generator.jl:47 [inlined]
  [6] Dict{String, Any}(kv::Base.Generator{DataFrames.DataFrameRows{DataFrame}, MLDatasets.var"#385#386"})
    @ Base ./dict.jl:103
  [7] read_ogb_metadata(fullname::String, dir::Nothing)
    @ MLDatasets ~/MLDatasets.jl/src/datasets/graphs/ogbdataset.jl:144
  [8] OGBDataset(fullname::String; dir::Nothing)
    @ MLDatasets ~/MLDatasets.jl/src/datasets/graphs/ogbdataset.jl:121
  [9] OGBDataset(fullname::String)
    @ MLDatasets ~/MLDatasets.jl/src/datasets/graphs/ogbdataset.jl:121
 [10] top-level scope
    @ REPL[2]:1
```

I am not exactly sure that this is the correct/intended behaviour.

johnnychen94:
The "core" packages should still be loaded eagerly, otherwise, we'll hit a lot of world-age issues like #135.

closes #135

cc: @Dsantra92

### DIFF
--- a/src/MLDatasets.jl
+++ b/src/MLDatasets.jl
@@ -16,7 +16,7 @@ include("require.jl") # export @require
 # Use `@lazy import SomePkg` whenever the returned types are not its own types,
 # since for methods applied on the returned types we would encounter in world-age issues
 # (see discussion in  https://github.com/JuliaML/MLDatasets.jl/pull/128).
-# In the other case instead, use `require import SomePkg` to force 
+# In the other case instead, use `require import SomePkg` to force
 # the use to manually import.
 
 @require import JSON3="0f8b85d8-7281-11e9-16c2-39a750bddbf1"
@@ -25,14 +25,14 @@ include("require.jl") # export @require
 # @lazy import NPZ # lazy imported by FileIO
 @lazy import Pickle="fbb45041-c46e-462f-888f-7c521cafbc2c"
 @lazy import MAT="23992714-dd62-5051-b70f-ba57cb901cac"
-@lazy import CSV="336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
+import CSV
 @lazy import HDF5="f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 # @lazy import JLD2
 
 export getobs, numobs # From MLUtils.jl
 
 include("abstract_datasets.jl")
-# export AbstractDataset, 
+# export AbstractDataset,
 #        SupervisedDataset
 
 include("utils.jl")


johnnychen94:
version resolving failure on nightly is unrelated https://github.com/oxinabox/DataDeps.jl/issues/154

CarloLucibello:
I guess this cannot be avoided unless we import whatever type can end up in CSV.File, which is probably `InlineStrings, 
PooledArrays, SentinelArrays, WeakRefStrings` based on the CSV.jl project file, but this doesn't seem much robust. 
Another alternative is to `@require CSV`, but I don't want to have another breaking change, so let's go with this.

CarloLucibello (approved):

Issue #139, opened 2022-05-28
Support for Heterogeneous Graphs


Dsantra92:
We should support [Heterogeneous graphs](https://pytorch-geometric.readthedocs.io/en/latest/notes/heterogeneous.html#:~:text=Heterogeneous%20graphs%20come%20with%20different,differences%20in%20type%20and%20dimensionality.).

Dsantra92:
Completed in #140

Issue #107, opened 2022-03-27
load torch tensors in OGBDatasets


CarloLucibello:
Some of the features of the OGBDataset are downloaded as torch tensor stored in the ".pt" format. They are currently ignored at the moment, but we could load them using Pickle.jl (e.g. see this [comment](https://github.com/FluxML/Metalhead.jl/issues/112#issuecomment-1074176899))

Dsantra92:
Can I work on this?

CarloLucibello:
Sure. I don't remember for which specific dataset this was needed though

CarloLucibello:
This problem can be seen in `OGBDataset("ogbl-collab")`

Dsantra92:
Been inactive due to Uni. exams, will start working on it today.

yuehhua:
Some problems have been overcome here, including loading ".pt" format using Pickle.jl and have been discussed with @chengchingwen : https://github.com/yuehhua/GraphMLDatasets.jl/blob/65d6a2bb02d31569a64b47004a0c4b192739a066/src/preprocess.jl#L391
Hope these code help.

PR #140, opened 2022-06-05
Support for OGB Hetero-Graphs

Dsantra92:
Handling Hetero Graphs in OGBDatasets.

### DIFF
--- a/src/graph.jl
+++ b/src/graph.jl
@@ -185,7 +185,7 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", d::HeteroGraph)
     recur_io = IOContext(io, :compact => false)
-    print(io, "HeteroGraph:")
+    print(io, "Heterogeneous Graph:")
     for f in fieldnames(HeteroGraph)
         if !startswith(string(f), "_")
             fstring = leftalign(string(f), 12)
--- a/src/graph.jl
+++ b/src/graph.jl
@@ -137,11 +137,11 @@ Nodes are indexed in `1:num_nodes`.
 - `num_edges`: Dictionary containing the number of edges for each relation.
 - `edge_indices`: Dictionary containing the `edge_index` for each edge relation. An `edge_index` is a tuple containing two vectors with length equal to the number of edges for the relation.
     The first vector contains the list of the source nodes of each edge, the second contains the target nodes.
-- `node_data`: node-related data. Can be `nothing`, a named tuple of arrays or a dictionary of arrays.
-            The array's last dimension size should be equal to the number of nodes.
+- `node_data`: node-related data. Can be `nothing`, Dictionary of a dictionary of arrays. Data of a speific type of node can be accessed 
+            using node_data[node_type].The array's last dimension size should be equal to the number of nodes.
             Default `nothing`.
-- `edge_data`: edge-related data. Can be `nothing`, a named tuple of arrays or a dictionary of arrays.
-            The array's last dimension size should be equal to the number of edges.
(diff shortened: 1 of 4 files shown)


CarloLucibello:
Related to #139

CarloLucibello:
what's the status of this? on which specific ogb dataset are you testing?

Dsantra92:
Testing it on `ogbn-mag`. Edge data section and docs is left. Completing by EOD.

Dsantra92:
Not adding training indexes in this pr. Training splits will be handled as per the [torch issue](https://github.com/JuliaML/MLDatasets.jl/issues/107#issuecomment-1154163832). We also need more documentation for HeteroGraphs (and maybe a tutorial). Not sure why docs are failing :)

src/graph.jl
```diff
@@ -115,10 +116,89 @@ function Base.show(io::IO, ::MIME"text/plain", d::Graph)
...
+struct HeteroGraph <: AbstractGraph
+    num_nodes::Dict{String, Int}
+    num_edges::Dict{Tuple{String, String, String}, Int}
+    edge_indices::Dict{Tuple{String, String, String}, Tuple{Vector{Int}, Vector{Int}}}
+    node_data
+    edge_data
+end
+
+function HeteroGraph(;
+    num_nodes::Union{Dict{String, Int}, Nothing}=nothing,
```
CarloLucibello, line 156: do we need `nothing`? can't we just use `Dict{String, Int}()` as default?

CarloLucibello:
Is ` ogbn-mag` big? can we add it to the test set? If not in the CI tests we should at least add it to the local tests

src/graph.jl
Dsantra92: Assigning `num_nodes` `Dict{String, Int}()` might indicate that there is no node in the Graph. Whereas `nothing` indicates that `num_nodes` information is not provided but can be inferred from other values.

test/datasets/graphs.jl
```diff
@@ -122,3 +122,26 @@ end
...
+            @test g.num_nodes[node_type] == size(features)[end]
+        end
+    end
+    # check if the number of edges are consistent
+    for edge_data in g.edge_data
+        for (edge_type, features) in edge_data
+            @test g.num_edges[edge_type] == size(features)[end]
+        end
+    end
+end
```
CarloLucibello, line 147: can you test also the sizes and num_nodes / num_edges specific to this dataset?

CarloLucibello:
`node_data` should be such that `node_data[node_type]` is a dict or a named tuples. Similar applies to edge_data.

src/graph.jl
```diff
@@ -115,10 +116,89 @@ function Base.show(io::IO, ::MIME"text/plain", d::Graph)
...
+- `edge_data`: edge-related data. Can be `nothing`, a named tuple of arrays or a dictionary of arrays.
+            The array's last dimension size should be equal to the number of edges.
+            Default `nothing`.
+"""
+struct HeteroGraph <: AbstractGraph
+    num_nodes::Dict{String, Int}
+    num_edges::Dict{Tuple{String, String, String}, Int}
+    edge_indices::Dict{Tuple{String, String, String}, Tuple{Vector{Int}, Vector{Int}}}
+    node_data
+    edge_data
```
CarloLucibello, line 152: should we also add the fields `node_types` and `edge_types`?

Dsantra92: that would also be nice to have

test/datasets/graphs.jl
Dsantra92: Done!!

PR #143, opened 2022-06-19
Use Manual deps for registering some datahubs

Dsantra92:
Things that will not work:

```julia
using MLDatasets, DataDeps

# before using a TUDataset
datadep"TUDataset" # tells the user to copy the required files to a datadeps folder
```

We will be supporting users to place their non-public datasets, so overriding internal functions for `ManualDataDeps` was not a viable option.


Dsantra92:
I don't understand why tests are failing on Windows :0

CarloLucibello:
Maybe we should exclude the ogb-mag test if Sys.iswindows()

PR #146, opened 2022-06-23
Add ImageNet

### LINKED ISSUE #100: Feature request: ImageNet data loader
ImageNet is quite large and locked behind [terms of access that require an account](https://www.image-net.org/download.php).

However it would be nice to be able to either
* set a config (or `ENV`) variable to download ImageNet through MLDatasets
* point MLDatasets to a local copy of ImageNet

and be able to use MLDatasets' interface of
```julia 
train_x, train_y = ImageNet.traindata()
test_x,  test_y  = ImageNet.testdata()
```
as well as `ImageNet.convert2image(x)`. 
Ideally data would be in WHCN format for Flux and Metalhead models.

adrhill:
Draft PR to add the ImageNet 2012 Classification Dataset (ILSVRC 2012-2017) as a `ManualDataDep`.
Closes #100.

___
Since ImageNet is very large (>150 GB) and [requires signing up and accepting the terms of access](https://image-net.org/), it can only be added manually. The `ManualDataDep` instruction message for ImageNet includes the following:
* instructions on creating symlinks to existing ImageNet datasets (e.g. for use on shared compute clusters)
* instructions on downloading and unpacking ImageNet from scratch, based on [the PyTorch guide](https://github.com/pytorch/examples/tree/main/imagenet), as [linked by @CarloLucibello](https://github.com/JuliaML/MLDatasets.jl/issues/100#issuecomment-1117053093)

When unpacked "PyTorch-style", the ImageNet dataset is assumed to look as follows: `ImageNet -> split-folder -> WordNet ID folder -> class samples as jpg-files`, e.g.:
```
ImageNet
├── train
├── val
│   ├── n01440764
│   │   ├── ILSVRC2012_val_00000293.JPEG
│   │   ├── ILSVRC2012_val_00002138.JPEG
│   │   └── ...
│   ├── n01443537
│   └── ...
├── test
└── devkit
    ├── data
    │   ├── meta.mat
    │   └── ...
    └── ...
```

## Current limitations
Since ImageNet is too large to precompute all preprocessed images and keep them in memory, the dataset precomputes a list of all file paths instead. ~~Calling `Base.getindex(d::ImageNet, i)` loads the image via ImageMagick.jl and preprocesses it when required. This adds dependencies on ImageMagick and Images.jl via LazyModules.~~

This also means that the `ImageNet` struct currently doesn't contain `features` (which might be a requirement for `SupervisedDataset`s?)


lorenzoh:
Will be good to have ImageNet support!

I'm wondering if there may be a simpler implementation for this, though. It seems the dataset has the same format as the (derived) ImageNette and ImageWoof datasets. The way those are loaded in FastAI.jl combines the MLUtils.jl primitives and those could be used to load ImageNet as folows:

```julia
using MLDatasets, MLUtils, FileIO

function ImageNet(dir)
    files = FileDataset(identity, path, "*.JPEG").paths
    return mapobs((FileIO.load, loadlabel), files)
end
# get the class name from the file path. could add a lookup here to convert the ID to the human-readable name
loadlabel(file::String) = split(file, "/")[end-2]


data  = ImageNet(IMAGENET_DIR)

# only training set
data  = ImageNet(joinpath(IMAGENET_DIR, "train"))
```

I'd also suggest using FileIO.jl for loading images which will the faster JpegTurbo.jl to load the images.

If more control over the image loading is desired, like converting to a color upon reading or loading an image into a smaller size (much faster if it'll be downsized during training anyway) , one could also use JpegTurbo.jl directly:

```julia
function ImageNet(dir; C = RGB{N0f8}, preferred_size = nothing)
    files = FileDataset(identity, path, "*.JPEG").paths
    return mapobs((f -> JpegTurbo.jpeg_decode(C, f; preferred_size), loadlabel), files)
end

# load as grayscale and smaller image size
data = ImageNet(IMAGENET_DIR; C = Gray{N0f8}, preferred_size = (224, 224))
```

adrhill:
Thanks a lot, loading smaller images with JpegTurbo is indeed much faster!

adrhill:
JpegTurbo's `preferred_size` keyword already returns images pretty close to the desired `224x224` size. At the cost of losing a couple of pixels, we could skip the second resizing in `resize_smallest_dimension`, which allocates and directly `center_crop`, which is just a view.

adrhill:
I've done some local benchmarks:
Current commit [cac14d2](https://github.com/JuliaML/MLDatasets.jl/pull/146/commits/cac14d25d1d102dd70020d6add47a0017dcbf535) with JpegTurbo loading smaller images:
```julia
julia> using MLDatasets

julia> dataset = ImageNet(Float32, :val);

julia> @benchmark dataset[1:16]
BenchmarkTools.Trial: 44 samples with 1 evaluation.
 Range (min … max):  104.413 ms … 143.052 ms  ┊ GC (min … max):  7.28% … 18.57%
 Time  (median):     113.164 ms               ┊ GC (median):    10.80%
 Time  (mean ± σ):   115.515 ms ±   9.030 ms  ┊ GC (mean ± σ):  10.46% ±  3.68%

    ▃          █                                                 
  ▇▄█▄▁▁▄▄▇▇▇▄▄█▄▇▄▄▇▁▄▁▄▁▁▄▁▄▄▁▄▁▁▄▄▁▁▁▄▁▁▁▁▁▄▁▄▁▁▁▄▁▁▁▁▁▁▁▁▁▄ ▁
  104 ms           Histogram: frequency by time          143 ms <

 Memory estimate: 131.78 MiB, allocs estimate: 2050.
```

Without `resize_smallest_dimension `, using only `center_crop`:
```julia
julia> @benchmark dataset[1:16]
BenchmarkTools.Trial: 57 samples with 1 evaluation.
 Range (min … max):  80.594 ms … 103.226 ms  ┊ GC (min … max):  7.43% … 19.03%
 Time  (median):     86.954 ms               ┊ GC (median):     8.95%
 Time  (mean ± σ):   88.287 ms ±   5.683 ms  ┊ GC (mean ± σ):  10.90% ±  3.57%

    ▄ ▄ ▁▁ █▄  ▁▄   ▁    ▁  ▁▁ ▁   ▄   ▁ ▁   ▁           ▁      
  ▆▆█▆█▁██▁██▆▁██▆▁▆█▁▁▆▁█▁▆██▆█▁▁▁█▆▁▆█▁█▁▁▁█▁▁▁▆▁▁▁▁▁▁▁█▁▁▁▆ ▁
  80.6 ms         Histogram: frequency by time          101 ms <

 Memory estimate: 115.96 MiB, allocs estimate: 1826.
```

Additionally using StackedViews.jl for batching:
```julia
julia> @benchmark dataset[1:16]
BenchmarkTools.Trial: 95 samples with 1 evaluation.
 Range (min … max):  47.971 ms … 73.503 ms  ┊ GC (min … max): 0.00% … 8.68%
 Time  (median):     51.116 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   52.903 ms ±  4.922 ms  ┊ GC (mean ± σ):  4.69% ± 5.81%

  ▂ ▂▄█                                                        
  █▇█████▃▅▃▅▅▆▃█▇▇▅▅▅▁▅▁▁▃▃▆▆▁▁▁▁▁▅▃▃▁▃▁▁▁▁▁▁▁▁▁▃▁▁▁▃▁▁▁▁▁▁▃ ▁
  48 ms           Histogram: frequency by time          70 ms <

 Memory estimate: 38.43 MiB, allocs estimate: 1499.
```

src/datasets/vision/imagenet_reader/ImageNetReader.jl
```diff
@@ -0,0 +1,48 @@
...
+readdata(dir::AbstractString) = FileDataset(identity, dir, "*.JPEG").paths
+
+# Get WordNet ID from path
+function load_wnids(files::AbstractVector{<:AbstractString})
+    return [split(f, "/")[end - 1] for f in files]
+end
+
+# Load image from ImageNetFile path and preprocess it to normalized 224x224x3 Array{Tx,3}
+function readimage(Tx::Type{<:Real}, file::AbstractString)
+    im = JpegTurbo.jpeg_decode(ImageCore.RGB{Tx}, file; preferred_size=IMGSIZE)
```
johnnychen94, line 38: I'm not sure if all ImageNet images meets the requirement, but note that the actual decomposed result size `size(im)` might not be `preferred_size`.

src/datasets/vision/imagenet_reader/preprocess.jl
```diff
@@ -0,0 +1,32 @@
...
+
+# Coefficients taken from PyTorch's ImageNet normalization code
+const PYTORCH_MEAN = [0.485f0, 0.456f0, 0.406f0]
+const PYTORCH_STD = [0.229f0, 0.224f0, 0.225f0]
+
+function preprocess(Tx::Type, im::AbstractMatrix{<:ImageCore.AbstractRGB})
+    im = center_crop_view(im)
+    im = (ImageCore.channelview(im) .- PYTORCH_MEAN) ./ PYTORCH_STD
+    # Convert from CHW (Image.jl's channel ordering) to WHC:
+    return Tx.(PermutedDimsArray(im, (3, 2, 1)))
```
johnnychen94, line 23: If we want to lazify every step, then MappedArrays.jl might be helpful here and to `inverse_preprocess`.

src/datasets/vision/imagenet.jl
```diff
@@ -0,0 +1,204 @@
...
+end
+
+function convert2image(::Type{<:ImageNet}, x::AbstractArray{<:Integer})
+    return convert2image(ImageNet, reinterpret(N0f8, convert(Array{UInt8}, x)))
+end
+convert2image(::Type{<:ImageNet}, x) = ImageNetReader.inverse_preprocess(x)
+
+Base.length(d::ImageNet) = length(d.image_files)
+function Base.getindex(d::ImageNet, ::Colon)
+    # Throw warning here that ImageNet probably will not fit in memory?
```
johnnychen94, line 199: If this almost certainly hits out-of-memory error, then we should probably just error eagerly, e.g., `Base.getindex(d::ImageNet, ::Colon) = error("Loading all images net images into memory is problematic.")`

src/MLDatasets.jl
```diff
@@ -29,6 +29,11 @@ import CSV
 @lazy import HDF5="f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 # @lazy import JLD2
 
+# Required for ImageNet
+@lazy import JpegTurbo="b835a17e-a41a-41e7-81f0-2f016b05efe0"   # Open jpg-files
+@lazy import ImageCore="a09fc81d-aa75-5fe9-8630-4744c3626534"   # Preprocessing
+@lazy import StackViews="cae243ae-269e-4f55-b966-ac2d0dc13c15"  # Batching of images
```
johnnychen94, line 35: I'm quite confident that this will hit the world-age issue. It's a very thin package so there's no need to worry about its latency.

```suggestion
import StackViews
```

src/datasets/vision/imagenet_reader/ImageNetReader.jl
```diff
@@ -0,0 +1,48 @@
+module ImageNetReader
+import ..FileDataset
+import ..read_mat
+import ..@lazy
+
+@lazy import JpegTurbo = "b835a17e-a41a-41e7-81f0-2f016b05efe0"
+@lazy import ImageCore="a09fc81d-aa75-5fe9-8630-4744c3626534"
+@lazy import StackViews="cae243ae-269e-4f55-b966-ac2d0dc13c15"
```
johnnychen94, line 8:
```suggestion
import StackViews
```

johnnychen94 (commented):
I'm against `@lazy import ImageCore` (and `@lazy import ImageShow`) because it's very likely to hit the world-age issue if not used carefully. I mean, if this is a safe solution I'll be the first one to refactor the JuliaImages ecosystem this way. But since @CarloLucibello is the actual maintainer of this package, I'll leave this decision to him.

src/datasets/vision/imagenet.jl
adrhill: Added `ArgumentError` in 02e966d with a message to use `dataset[1:end]` if loading all of the dataset is indeed the desired behaviour.

src/datasets/vision/imagenet_reader/ImageNetReader.jl
johnnychen94: The reason for this is that JpegTurbo.jl (or libjpegt-turbo) only supports a very limited range of `scale_ratio`: they are $M/8$ where $M  \in [1, 2, ..., 16]$. Thus the maximal possible `scale_ratio` is 2.

The supported `scale_ratio` permits a faster decoding algorithm (by scaling the coefficients instead of the actual images), this is why we can observe the performance boost here.

---

The perhaps safest (_I think_) solution is to add a `imresize` after it:

```julia
    img = @suppress_err JpegTurbo.jpeg_decode(file; preferred_size=(224, 224))
    if size(img) != (224, 224)
        img = imresize(img, (224, 224))
    end
```

The `@suppress_err` macro is a handy tool from https://github.com/JuliaIO/Suppressor.jl to disable this warning message.

src/datasets/vision/imagenet_reader/ImageNetReader.jl
```diff
@@ -0,0 +1,47 @@
+module ImageNetReader
+using ImageCore: channelview, colorview, AbstractRGB, RGB
+
+import ..FileDataset
+import ..read_mat
+import ..@lazy
+
+@lazy import JpegTurbo = "b835a17e-a41a-41e7-81f0-2f016b05efe0"
```
Dsantra92, line 8: This file is called in MLDatasets.jl, are these imports necessary?

adrhill:
Thanks for the review @Dsantra92! 
I'm slightly busy due to the JuliaCon submission deadline on Monday, but I'll get back to this PR as soon as possible.

src/datasets/vision/imagenet.jl
```diff
@@ -0,0 +1,198 @@
...
+            │   ├── n01443537
+            │   └── ...
+            ├── test
+            └── devkit
+                ├── data
+                │   ├── meta.mat
+                │   └── ...
+                └── ...
+            ```
+            If your existing copy of the ImageNet dataset uses another file structure,
```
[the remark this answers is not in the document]
adrhill: Should we host it in a separate `.sh` file somewhere in the repo and link to that? That's how it's currently done in the PyTorch examples: https://github.com/pytorch/examples/blob/d5478765d38210addf474dd73faf0d103052027a/imagenet/extract_ILSVRC.sh

src/datasets/vision/imagenet_reader/ImageNetReader.jl
adrhill: This is part of the `ImageNetReader` module inside MLDatasets. I thought this was necessary to import packages and functions from the parent module `MLDatasets`.

adrhill:
The order of the classes in the metadata also still has to be fixed as it doesn't match  https://raw.githubusercontent.com/pytorch/hub/master/imagenet_classes.txt.

src/datasets/vision/imagenet.jl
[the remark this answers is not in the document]
adrhill: I moved the instructions to `docs/src/datasets/imagenet_installation.md`. 
Do you want to host them in the docs or just directly link to the markdown file on GitHub?

src/datasets/vision/imagenet_reader/preprocess.jl
```diff
@@ -0,0 +1,31 @@
...
+    ]
+end
+
+# Coefficients taken from PyTorch's ImageNet normalization code
+const PYTORCH_MEAN = [0.485f0, 0.456f0, 0.406f0]
+const PYTORCH_STD = [0.229f0, 0.224f0, 0.225f0]
+
+function preprocess(Tx::Type, im::AbstractMatrix{<:AbstractRGB})
+    im = center_crop_view(im)
+    im = (channelview(im) .- PYTORCH_MEAN) ./ PYTORCH_STD
```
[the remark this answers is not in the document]
adrhill: I took another look at the PyTorch example. When creating a dataset, they allow a `transform` function to be passed, so the normalisation is indeed not part of the dataset:

https://github.com/pytorch/examples/blob/91ccd7a21be6fa687000beef82fc1e5d7d64e4bd/imagenet/main.py#L211-L230
```python
normalize = transforms.Normalize(mean=[0.485, 0.456, 0.406],
                                 std=[0.229, 0.224, 0.225])

train_dataset = datasets.ImageFolder(
    traindir,
    transforms.Compose([
        transforms.RandomResizedCrop(224),
        transforms.RandomHorizontalFlip(),
        transforms.ToTensor(),
        normalize,
    ]))

val_dataset = datasets.ImageFolder(
    valdir,
    transforms.Compose([
        transforms.Resize(256),
        transforms.CenterCrop(224),
        transforms.ToTensor(),
        normalize,
    ]))
```

More interestingly, all of the resizing and cropping also isn't "hardcoded" into the data loader.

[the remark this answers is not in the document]
CarloLucibello: Yes we should remove them

adrhill:
Sorry for stalling this.

I guess the issue with this PR boils down to whether preprocessing functions belong to MLDatasets or to packages exporting pre-trained models. This question has already been raised in https://github.com/FluxML/Metalhead.jl/issues/117.

Since images in ImageNet have different dimensions, providing an ImageNet data loader with matching preprocessing functions would be somewhat useless, as it would not be able to load batches of data. And as discussed here in the context of `JpegTurbo.jpeg_decode`, a lot of performance would be left on the table if we loaded full-size images just to immediately resize them. 

I took a look at how other Deep Learning frameworks deal with this and both [torchvision](https://pytorch.org/vision/stable/models.html#using-the-pre-trained-models) and [Keras Applications](https://keras.io/api/applications/) export preprocessing functions with their pre-trained models. MLDataset's `FileDataset` pattern would work well if pre-trained model libraries exported a corresponding `loadfn`.
One of the issues mentioned in https://github.com/FluxML/Metalhead.jl/issues/117 is import latency for extra dependencies such as DataAugmentation.jl. Maybe LazyModules.jl could help circumvent this problem.

RomeoV:
Hey everyone, this looks awesome. Is anyone still working on this? Otherwise I would suggest trying to merge this, even if it's not "perfect" with regards to extra dependencies or open questions about transformations.

src/datasets/vision/imagenet_reader/preprocess.jl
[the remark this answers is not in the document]
RomeoV: I do think having a standarized way to preprocess should be distributed with the DataSet, at least in the form of a Readme (e.g. by posting the code above), so that it is easy for people to get started.

adrhill:
I'm still interested in working on this. 

To get this merged, we could make the `preprocess` and `inverse_preprocess` functions part of the `ImageNet` struct and provide the current functions as defaults.

src/datasets/vision/imagenet_reader/preprocess.jl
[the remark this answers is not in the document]
adrhill: [09feb3d](https://github.com/JuliaML/MLDatasets.jl/pull/146/commits/09feb3d4a621df3fca6e38f3fb9119b6bcb9ac4b) adds this feature to the constructor:
```Julia
function ImageNet(
    Tx::Type,
    split::Symbol;
    preprocess=ImageNetReader.default_preprocess,
    inverse_preprocess=ImageNetReader.default_inverse_preprocess,
    # ...
)

**Comment by CarloLucibello** [comment]
this needs a rebase, otherwise looks mostly good

File: src/datasets/vision/imagenet.jl [hunk 9]
```diff
@@ -0,0 +1,155 @@
...
+            Please follow the instructions at
+            https://github.com/JuliaML/MLDatasets.jl/blob/master/docs/src/datasets/imagenet_installation.md.
+            """,
+        ),
+    )
+end
+
+"""
+    ImageNet(; Tx=Float32, split=:train, dir=nothing)
+    ImageNet([Tx, split])
```
CarloLucibello, line 21: there are many keyword arguments that need to be documented

src/datasets/vision/imagenet_reader/preprocess.jl
```diff
@@ -0,0 +1,34 @@
+# Image preprocessing for ImageNet models.
+# Code adapted from Metalhead 0.5.3's utils.jl
+
+# Coefficients taken from PyTorch's ImageNet normalization code
+const PYTORCH_MEAN = [0.485f0, 0.456f0, 0.406f0]
+const PYTORCH_STD = [0.229f0, 0.224f0, 0.225f0]
+
+normalize_pytorch(x) = (x .- PYTORCH_MEAN) ./ PYTORCH_STD
+inv_normalize_pytorch(x) = x .* PYTORCH_STD .+ PYTORCH_MEAN
```
ToucheSir, line 9: I would drop the `pytorch` prefix/suffix and use something else. The comments can stay. If PyTorch isn't the only library that does this preprocessing, then it makes sense to represent that with more general names. If different libraries are providing different preprocessing functionality for ImageNet (or not providing any), then I'd argue there is no canonical default set of ImageNet transformations and this code (aside from maybe the descriptive stats) shouldn't be in MLDatasets.

src/datasets/vision/imagenet.jl
adrhill: Updated in 944bd83.

src/datasets/vision/imagenet_reader/preprocess.jl
adrhill: Good point. Since this is just an internal function used by `default_preprocess`, I would suggest either `_normalize` or `default_normalize`. The appeal of using these coefficients as defaults is that they should work out of the box with pre-trained vision models from [Metalhead.jl](https://github.com/FluxML/Metalhead.jl/).

ToucheSir: Wait, so do other libraries provide this functionality in their ImageNet dataset APIs? I checked https://www.tensorflow.org/datasets/catalog/imagenet2012 and it has no mention of preprocessing, so is PyTorch the only library that does this? If so, I would vote to remove the preprocessing functions as mentioned above.

Dsantra92: If I am not wrong, these normalization values depend on the model you are using. Also,  none of the existing vision datasets have preprocessing functions. These functions are ideally handled by data preprocessing libraries/modules.

ToucheSir: The norm values should not be model-specific. They're derived directly from the data before any model is involved.

RomeoV: For reference, pytorch includes normalization constants as part of the "model weights" ([see here](https://pytorch.org/vision/stable/models/generated/torchvision.models.resnet34.html#torchvision.models.ResNet34_Weights)):
> The inference transforms are available at ResNet34_Weights.IMAGENET1K_V1.transforms and perform the following preprocessing operations: Accepts PIL.Image, batched (B, C, H, W) and single (C, H, W) image torch.Tensor objects. The images are resized to resize_size=[256] using interpolation=InterpolationMode.BILINEAR, followed by a central crop of crop_size=[224]. Finally the values are first rescaled to [0.0, 1.0] and then normalized using mean=[0.485, 0.456, 0.406] and std=[0.229, 0.224, 0.225].

Another famous pretrained-models library, the [timm library](https://github.com/rwightman/pytorch-image-models/blob/main/timm/data/constants.py), includes the normalization weights as `from timm.data.constraints import IMAGENET_DEFAULT_MEAN, IMAGENET_DEFAULT_STD`.

RomeoV: In the pytorch case, notice however that although the transformations are stored in the "model weights", the mean and std is the same across models ([see e.g. the mobilenet model](https://pytorch.org/vision/stable/models/generated/torchvision.models.mobilenet_v3_large.html#torchvision.models.MobileNet_V3_Large_Weights)).

In a similar spirit, I would definitely defend the decision of shipping the set of transformations (cropping, interpolation, linear transformation, etc) as part of the dataset. However I agree with the very first point that the name `transformation_pytorch` isn't really precise, although I think it is fair to link to the corresponding transformations for tensorflow, pytorch, and/or the timm library in a related comment.

ToucheSir: PyTorch also lumps code for pretrained models, data augmentations and datasets into one library, I don't think we need to follow their every example :)

> In a similar spirit, I would definitely defend the decision of shipping the set of transformations (cropping, interpolation, linear transformation, etc) as part of the dataset.

This is precisely why I asked about what other libraries are doing. If nobody else is shipping the same set of transformations, then they can hardly be considered canonical for ImageNet. That doesn't mean we should never ship helpers to create common augmentation pipelines, but that is better served by packages which have access to efficient augmentation libraries (e.g. Augmentor, DataAugmentation) and not by some unoptimized implementation which is simultaneously more general (because it's applicable to other datasets) and less general (because many papers using ImageNet do not use these augmentations) than the dataset it's been attached to.

CarloLucibello: Let's just apply the `channelview` and permute transformation by default here,
and make the (permuted) mean and std values be part of the type

src/datasets/vision/imagenet.jl
```diff
@@ -0,0 +1,171 @@
...
+Base.length(d::ImageNet) = length(d.dataset)
+
+const IMAGENET_MEM_WARNING = """Loading the entire ImageNet dataset into memory might not be possible.
+    If you are sure you want to load all of ImageNet, use `dataset[1:end]` instead of `dataset[:]`.
+    """
+Base.getindex(::ImageNet, ::Colon) = throw(ArgumentError(IMAGENET_MEM_WARNING))
+Base.getindex(d::ImageNet, i::Integer) = (features=d.dataset[i], targets=d.targets[i])
+function Base.getindex(d::ImageNet, is::AbstractVector)
+    return (features=StackView(d.dataset[is]), targets=d.targets[is])
+end
```
CarloLucibello, line 171:
```suggestion
function Base.getindex(d::ImageNet, is::AbstractVector)
    return (features=StackView(d.dataset[is]), targets=d.targets[is])
end
```
Let's remove this method entirely and avoid the `StackView` dependency. Collation can be performed by downstream consumers, e.g. dataloaders

src/datasets/vision/imagenet_reader/ImageNetReader.jl
```diff
@@ -0,0 +1,48 @@
...
+
+const NCLASSES = 1000
+const IMGSIZE = (224, 224)
```
CarloLucibello, line 11: these two could be part of metadata

src/datasets/vision/imagenet_reader/preprocess.jl
adrhill: I have also taken a look at Keras' ImageNet utilities. While these normalization constants are used in many places throughout [torchvision](https://github.com/search?q=repo%3Apytorch%2Fvision%200.406&type=code) and [PyTorch](https://github.com/search?q=repo%3Apytorch%2Fpytorch+0.406&type=code), it looks like TensorFlow and Keras do indeed [use their own constants](https://github.com/keras-team/keras/blob/db11e9674157c9125c9215b3bd4928675cf5af53/keras/applications/imagenet_utils.py#L200-L219).

I agree with @ToucheSir's sentiment
> If nobody else is shipping the same set of transformations, then they can hardly be considered canonical for ImageNet.

However, this point can be drawn even further, as nothing about ImageNet is truly canonical.
To give some examples (some of which have previously been discussed):

1) There is no canonical reason why Images have to be loaded in 224 x 224 format.
2) There is no canonical reason to apply the resizing algorithm JpegTurbo.jl uses when calling `jpeg_decode` with a `preferred_size`.
3) There is no canonical way of sorting class labels.  Some sort by WordNet ID (e.g. PyTorch), others don't.

### Getting this merged
So make this dataloader as "unopinionated" as possible, we could just make it a very thin wrapper around `FileDataset` which just loads metadata. This would require the user to pass a `loadfn` which handles the transformation from file path to array. Class ordering could be handled using a `sort_by_wnid=True`  keyword argument and all new dependencies introduced in this PR could  be removed (ImageCore, JpegTurbo and StackViews).

### Future work
However, I do strongly feel like some package in the wider Julia ML / Deep Learning ecosystem should export `loadfn`s that are usable with Metalhead's PyTorch models out of the box. @lorenzoh previously proposed adding such functionality to DataAugmentation.jl in https://github.com/FluxML/Metalhead.jl/issues/117. 
Once this functionality is available somewhere, `ImageNet`'s docstring in MLDataset should be updated to showcase this common use-case.

Until this functionality exists, I would suggest adding a `"Home" => "Tutorials" => "ImageNet"` page to the MLDatasets docs which implements the current load function.

ToucheSir: > it looks like TensorFlow and Keras do indeed [use their own constants](https://github.com/keras-team/keras/blob/db11e9674157c9125c9215b3bd4928675cf5af53/keras/applications/imagenet_utils.py#L200-L219).

Nice find. I was not expecting that `mode == "torch"` conditional.

> However, this point can be drawn even further, as nothing about ImageNet is truly canonical. To give some examples (some of which have previously been discussed):
> 
>     1. There is no canonical reason why Images have to be loaded in 224 x 224 format.
> 
>     2. There is no canonical reason to apply the resizing algorithm JpegTurbo.jl uses when calling `jpeg_decode` with a `preferred_size`.
> 
>     3. There is no canonical way of sorting class labels.  Some sort by WordNet ID (e.g. PyTorch), others don't.

The difference here is that all three of those points can have a decent fallback without depending on external packages. Another argument is that more people will rely on these defaults than won't. I'm not sure augmentations pass that threshold.
I'm not saying users shouldn't be able to pass in a transformation function, but `identity` or some such seems a more defensible default. Indeed, the torchvision `ImageNet` class does not do any additional transforms by default, so we'd be deviating from every other library if we stuck with this default centre crop.

src/datasets/vision/imagenet.jl
adrhill: Removed in fe38d43.

src/datasets/vision/imagenet_reader/ImageNetReader.jl
adrhill: I've added `metadata["n_classes"]`. Should we make `imgsize` an argument to the `ImageNet` constructor or lock it to 224 x 224?

src/datasets/vision/imagenet_reader/preprocess.jl
adrhill: Resolving this due to the removal of normalization in the default preprocessing function.

src/datasets/vision/imagenet_reader/ImageNetReader.jl
ToucheSir: That would be great. What happens if we don't pass `preferred_size` to JpegTurbo?

adrhill: According to https://github.com/JuliaML/MLDatasets.jl/pull/146#issuecomment-1164557500, this would slow down the image loading:
> If more control over the image loading is desired, like converting to a color upon reading or loading an image into a smaller size (much faster if it'll be downsized during training anyway) , one could also use JpegTurbo.jl directly

src/datasets/vision/imagenet.jl
```diff
@@ -0,0 +1,186 @@
...
+    split::Symbol;
+    img_size::Tuple{Int,Int}=(224, 224),
+    preprocess=ImageNetReader.default_preprocess,
+    inverse_preprocess=ImageNetReader.default_inverse_preprocess,
+    dir=nothing,
+    train_dir="train",
+    val_dir="val",
+    test_dir="test",
+    devkit_dir="devkit",
+)
```
CarloLucibello, line 132:
```suggestion
        Tx::Type,
        split::Symbol;
        img_size::Tuple{Int,Int}=(224, 224),
        preprocess=ImageNetReader.default_preprocess,
        inverse_preprocess=ImageNetReader.default_inverse_preprocess,
        dir=nothing,
        train_dir="train",
        val_dir="val",
        test_dir="test",
        devkit_dir="devkit",
    )
```

src/datasets/vision/imagenet.jl
```diff
@@ -0,0 +1,186 @@
...
+julia> X, y = dataset[2000];
+
+julia> convert2image(dataset, X)
+
+julia> dataset.metadata
+Dict{String, Any} with 8 entries:
+  "features_dir"      => "/Users/funks/.julia/datadeps/ImageNet/val"
+  "class_WNIDs"       => ["n01440764", "n01443537", "n01484850", "n01491361", "n01494475", "
+  "class_description" => ["freshwater dace-like game fish of Europe and western Asia noted f
+  "n_observations"    => 50000
```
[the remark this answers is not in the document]
CarloLucibello: ah this is validation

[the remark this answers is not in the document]
adrhill: Exactly, 50k for the validation set:
```Julia
TRAINSET_SIZE = 1_281_167
VALSET_SIZE = 50_000
TESTSET_SIZE = 100_000
```

src/datasets/vision/imagenet.jl
adrhill: If you have a preferred code style, I can run JuliaFormatter on all new files.

adrhill:
In case someone is still interested in using this, I've opened a unregistered repository containing this PR: 
https://github.com/adrhill/ImageNetDataset.jl

The most notable difference is that ImageNetDataset.jl contains some custom preprocessing pipelines that support `convert2image` and work out of the box with Metalhead.jl.

adrhill:
Would anyone mind if I register [ImageNetDataset.jl](https://github.com/Julia-XAI/ImageNetDataset.jl) and close this PR?

ToucheSir:
No objections from me.

adrhill:
Closing with the registration of ImageNetDataset.jl:
https://github.com/JuliaRegistries/General/pull/126353

PR #150, opened 2022-07-01
Fix documentation typos for vision datasets

christiangnrd:
An extra colon seems to have snuck into the documentation string for most of the vision datasets.

That is all.


christiangnrd:
It seems like the tests that failed are unrelated to this pull request.

CarloLucibello:
@Dsantra92 looks like ml-1m is occasionally failing (slow HTTP response?), maybe we can exclude it from CI

CarloLucibello:
@christiangnrd thanks for this PR!

Issue #151, opened 2022-07-01
New test failing ("ml-10m")


christiangnrd:
I'm getting a really strange error in a local test that I can't seem to figure out why.

I've attached a screenshot, but it says that there's missing "end" with the "begin" at line 115 of the graphs_no_ci.jl test file, but the "end" seems to be there.

I'd like to know if I am the only one experiencing this. 

It happens on Julia 1.7.3 running Ubuntu 22.04.
I also tested it on Windows 11 and the same error appeared.

![Screenshot from 2022-07-01 19-47-14](https://user-images.githubusercontent.com/28689358/176975397-20874d0c-15a0-4fdf-9812-e02b6b1f7f8a.png)

Dsantra92:
Let me look into this.

christiangnrd:
Hi @Dsantra92, the patch fixed the original error, but now I'm getting a new error.
`     Testing Running tests...
[ Info: Testing larger datasets
graphs_no_ci.jl: Error During Test at /home/christian/.julia/packages/MLDatasets/fdW2b/test/runtests.jl:48
  Got exception outside of a @test
  LoadError: syntax: incomplete: "begin" at /home/christian/.julia/packages/MLDatasets/fdW2b/test/datasets/graphs_no_ci.jl:115 requires end
  Stacktrace:
    [1] top-level scope
      @ ~/.julia/packages/MLDatasets/fdW2b/test/datasets/graphs_no_ci.jl:115
    [2] include(fname::String)
      @ Base.MainInclude ./client.jl:451
    [3] macro expansion
      @ ~/.julia/packages/MLDatasets/fdW2b/test/runtests.jl:49 [inlined]
    [4] macro expansion
      @ ~/.julia/juliaup/julia-1.7.3+0~x64/share/julia/stdlib/v1.7/Test/src/Test.jl:1359 [inlined]
    [5] macro expansion
      @ ~/.julia/packages/MLDatasets/fdW2b/test/runtests.jl:48 [inlined]
    [6] macro expansion
      @ ~/.julia/juliaup/julia-1.7.3+0~x64/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
    [7] top-level scope
      @ ~/.julia/packages/MLDatasets/fdW2b/test/runtests.jl:42
    [8] include(fname::String)
      @ Base.MainInclude ./client.jl:451
    [9] top-level scope
      @ none:6
   [10] eval
[...]

PR #154, opened 2022-07-02
Patch/fix ci

Dsantra92:
Fix CI issues mentioned in #151 #150


Dsantra92:
There seems to be some problem with the link itself. I will check if the link is working fine.

CarloLucibello:
Let's put all of the movielens tests in the no_ci file for the time being and merge this as quickly as possible since having broken CI is problematic for other PRs

Dsantra92:
> Let's put all of the movielens tests in the no_ci file for the time being and merge this as quickly as possible since having broken CI is problematic for other PRs

Okay, that seems to be a better option

Issue #167, opened 2022-09-01
World age issue wtih read_pytorch


Dsantra92:
`read_pytorch` throws world age issues.

```julia-repl
julia> MLDatasets.read_pytorch("train.pt")
ERROR: MethodError: no method matching THload(::String)
The applicable method may be too new: running in world age 32451, while current world is 32460.
Closest candidates are:
  THload(::AbstractString) at ~/.julia/packages/Pickle/pwvBM/src/torch/torch_load.jl:62 (method too new to be called from this world context.)
  THload(::Pickle.Torch.TorchPickler, ::Any) at ~/.julia/packages/Pickle/pwvBM/src/torch/torch_load.jl:66 (method too new to be called from this world context.)
Stacktrace:
 [1] read_pytorch(path::String)
   @ MLDatasets ~/MLDatasets.jl/src/io.jl:28
 [2] top-level scope
   @ REPL[2]:1
```

PR #168, opened 2022-09-01
Train test and validation for graph datasets. 

### LINKED ISSUE #158: Better Train, test, val split for graphs
## Problems

1. Splitting graphs is a bit more complicated than normal data. Graphs can be split based on node, edge or whole graph itself. We should be able to support all of that with minimal alterations. Currently, graph splits in MLDatases.jl have no specification.
2. The API for accessing train, test and val split is inconsistent.

## Proposed solution
I propose the following API:
1. Splits should be present in the metadata. We can access splits using `splits = dataset.metadata.split`
2. The split is a named tuple for train, test and val. Is empty if no predefined split is present.
3. Each split is a named tuple/struct which contains type of split $\in$ [node, edge, graph] and the `value` of the split.
4. There can be two or more splits depending on the feature you are splitting on. There are such examples in OGBDatasets. So it makes sense to make the `value` of the split a dict or named tuple.

This should ideally handle most of the cases. We only support non-dynamic graphs for now, I have not worked with dynamic graphs. Most likely we can support splits with little to no changes.

This recursive structure of splits can be confusing for the end-user. We can support APIs in MLUtils.jl like:

```julia
train_data, test_data, val_data = split(dataset)
# If no split is present in the dataset, all data will training data or error out.

train_data = split(dataset, :train)

# If we want to choose edge split
train_data = split(dataset, :train, :edge)

# Get training split for a feature
train_data = split(dataset, :train, :feature_name)
```
Keeping this a discussion for now.

cc: @CarloLucibello

Dsantra92:
Closes #158

Additionally #167

### DIFF
--- a/src/abstract_datasets.jl
+++ b/src/abstract_datasets.jl
@@ -4,7 +4,7 @@
 Super-type from which all datasets in MLDatasets.jl inherit.
 
 Implements the following functionality:
-- `getobs(d)` and `getobs(d, i)` falling back to `d[:]` and `d[i]` 
+- `getobs(d)` and `getobs(d, i)` falling back to `d[:]` and `d[i]`
 - Pretty printing.
 """
 abstract type AbstractDataset <: AbstractDataContainer end
@@ -19,9 +19,9 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", d::D) where D <: AbstractDataset
     recur_io = IOContext(io, :compact => false)
-    
+
     print(io, "dataset $(D.name.name):")  # if the type is parameterized don't print the parameters
-    
+
     for f in fieldnames(D)
         if !startswith(string(f), "_")
             fstring = leftalign(string(f), 10)
@@ -34,7 +34,7 @@ function Base.show(io::IO, ::MIME"text/plain", d::D) where D <: AbstractDataset
 end
 
 function leftalign(s::AbstractString, n::Int)
-    m = length(s) 
+    m = length(s)
     if m > n
         return s[1:n]
     else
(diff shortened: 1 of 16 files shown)


CarloLucibello:
Can you edit the OP and describe the changes in this PR?

> Additionally https://github.com/JuliaML/MLDatasets.jl/issues/167

additionally what?

src/abstract_datasets.jl
```diff
@@ -119,12 +117,23 @@ const ARGUMENTS_SUPERVISED_ARRAY = """
...
 - `dataset[:]`: Return all observations as a named tuple of features and targets.
 - `length(dataset)`: Number of observations.
 """
+
+"""
+    GraphDataset <: AbstractDataset
+
+An abstract dataset type for graph learning tasks.
+"""
+abstract type GraphDataset <: AbstractDataset end
```
CarloLucibello, line 135: better rename this as
```suggestion
abstract type AbstractGraphDataset <: AbstractDataset end
```
especially if we go with #169

CarloLucibello:
Let's split the fix to #167 to a separate PR

src/abstract_datasets.jl
```diff
@@ -19,9 +19,9 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", d::D) where D <: AbstractDataset
     recur_io = IOContext(io, :compact => false)
-
```
CarloLucibello, line 22: it would be good to avoid polluting PRs with this formatting changes, there should be some option in your editor to avoid doing this

src/abstract_datasets.jl
CarloLucibello: Also, it is better to file this change as a separate PR. Generally a PR should target a single issue or a single feature addition

Dsantra92:
Too many changes happened in this pr. I will be closing this pr and making separate prayer for each of them.

PR #170, opened 2022-09-01
require explicit Pickle import in load_pytorch

Dsantra92:
Moving `Pickle` to required packages. This means we have to load `Pickle` explicitly for Planetoid graph datasets.


CarloLucibello:
Since the world age problem is with the `read_pytorch` method, not used yet in the codebase, and not with `read_pickle`, can we try to fix #167 in a non-breaking way? 

We could try adding as the first line inside `load_pytorch` the following:
```julia
assert_imported(Pickle._lazy_pkgid)
```

Dsantra92:
`read_pytorch` will be used in OGB Dataset link splits.

There have only been breaking changes recently, so this fix is a good idea. Just to be sure, this is a one time hot-patch thing that we are revisiting later?

CarloLucibello:
> There have only been breaking changes recently, 

I don't think we have breaking changes since the last release, and in this case it seems pretty easy to avoid them.

Dsantra92:
@CarloLucibello this was required for #172

PR #172, opened 2022-09-03
Split for OGBDatasets 

Dsantra92:
The pr adds support for simple graph link split (#107) as well as support for splits in Heterogeneous OGB Datasets.

Need Fixing before merge: #167

### DIFF
--- a/src/datasets/graphs/ogbdataset.jl
+++ b/src/datasets/graphs/ogbdataset.jl
@@ -345,9 +345,22 @@ function read_ogb_graph(path, metadata)
 
         for key in keys(split_dict)
           if !isnothing(split_dict[key])
-            split_dict[key]["edge"] .+= 1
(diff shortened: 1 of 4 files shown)


Dsantra92:
Tests and some fixes are left.

Dsantra92:
The split for heterogeneous datasets with edge-level tasks needs to be addressed. Now that I have had the chance to think about it, I really don't want to change HeteroGraph API.

CarloLucibello:
Can we add a test with the dataset mentioned  in https://github.com/JuliaML/MLDatasets.jl/issues/107#issuecomment-1120384637?

> The split for heterogeneous datasets with edge-level tasks needs to be addressed. 

it can be done in a separate PR if it is too involved

Dsantra92:
Merge #170 before this

Dsantra92:
I think we are in a good position to merge this @CarloLucibello

Issue #169, opened 2022-09-01
Make an unifying type for all graph datasets


Dsantra92:
It would be good to have something like:

```julia
struct GraphDataset <: AbstractDataset
    name::String
    metadata::Dict{String, Any}
    graphs::Vector{Graph}
    graph_data::Union{Nothing,NamedTuple}
end
```
Instead of individual struct for each dataset. Unifying them would reduce a lot of code duplication.

Also, metadata fields can be adopted from the [OGB convention](https://github.com/snap-stanford/ogb/blob/master/ogb/linkproppred/master.csv). 

cc: @CarloLucibello

Dsantra92:
Closing this as not planned.

PR #173, opened 2022-09-10
Unify all Graph Dataset types

Dsantra92:
Related to: #169 

Added the GraphDataset struct which can replace all custom structs for each Data source.

The metadata will have fields like OGBDatasets metadata, including name, source, etc. We also can add data citations. This is also required for saving datasets in parquet format, as defining storing method for each struct is troublesome.

@CarloLucibello if you can review the initial struct, I can start with replacing the data source structs.

### DIFF
--- a/src/graph.jl
+++ b/src/graph.jl
@@ -227,3 +227,33 @@ function edgeindex2adjlist(s, t, num_nodes; inneigs=false)
     end
     return adj
 end
+
+"""
+    GraphDataset{<:AbstractGraph} <: AbstractDataset
+
+GraphDataset
+"""
+struct GraphDataset{T<:AbstractGraph} <: AbstractDataset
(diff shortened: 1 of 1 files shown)


CarloLucibello:
I'm not okay with replacing every graph dataset with a unique structure. At least we should first have a long discussion about that. 

Each dataset in MLDatasets used to be its own module, then with a large and breaking change datasets became types. I don't see enough compelling reasons to go through another heavily breaking change. 

Having datasets as their own types means we can customize their behavior with specialized methods, we don't want to lose that flexibility.

We can still expose a GraphDataset for the user's convenience in defining their own datasets, similarly to the containers in https://github.com/JuliaML/MLDatasets.jl/tree/master/src/containers

Dsantra92:
I didn't look into containers enough, I will look into that.

PR #183, opened 2022-10-21
Corrects docstring for the `as_df` keyword

vitaminace33:
The specific behavior of the `as_df` keyword stated in the docstring template for specific fields is contrary to the keyword's main description, the keyword meaning and to its actual behavior: `as_df=false` for plain arrays, `as_df=true` for dataframes.


CarloLucibello:
thanks!

vitaminace33:
yrw :D

Issue #201, opened 2023-01-11
Problem loading OGB dataset ogbg-ppa


chrisn-pik:
I can not load the OGBDataset: "ogbg-ppa"

```julia
using MLDatasets, DataFrames
data = OGBDataset("ogbg-ppa")
```
but get the following error:

```julia
ERROR: MethodError: no method matching parse_pystring(::Missing)
Closest candidates are:
  parse_pystring(::AbstractString) at ~/.julia/packages/MLDatasets/bg0uc/src/utils.jl:2
Stacktrace:
 [1] (::MLDatasets.var"#411#412")(r::DataFrameRow{DataFrame, DataFrames.Index})
   @ MLDatasets ./none:0
 [2] iterate
   @ ./generator.jl:47 [inlined]
 [3] Dict{String, Any}(kv::Base.Generator{DataFrames.DataFrameRows{DataFrame}, MLDatasets.var"#411#412"})
   @ Base ./dict.jl:105
 [4] read_ogb_metadata(fullname::String, dir::Nothing)
   @ MLDatasets ~/.julia/packages/MLDatasets/bg0uc/src/datasets/graphs/ogbdataset.jl:160
 [5] OGBDataset(fullname::String; dir::Nothing)
   @ MLDatasets ~/.julia/packages/MLDatasets/bg0uc/src/datasets/graphs/ogbdataset.jl:132
 [6] OGBDataset(fullname::String)
   @ MLDatasets ~/.julia/packages/MLDatasets/bg0uc/src/datasets/graphs/ogbdataset.jl:131
 [7] top-level scope
   @ REPL[8]:1

```

I am using Julia 1.8.2, DataFrames v1.3.6 and MLDatasets v0.7.9

chrisn-pik:
@CarloLucibello Thanks! I am not getting the error message anymore, however I run out of memory. I tried it on two computers, and it did not work. Are 16GB of RAM not enough to use this dataset? My computer crashes once the download is finished, when trying to extract the data.

PR #204, opened 2023-01-18
Handle missing values in OGB master csv

Dsantra92:
There are some missing value in OGB master CSVs. This pr aims to handle missing value for creating metadata.

Fix: #201


CarloLucibello:
can you add the simple test (either ci or no-ci, depends on the size of the dataset) checking that `data = OGBDataset("ogbg-ppa")` is loaded properly?

Dsantra92:
I am not sure if the dataset will fit my computer memory. I will try none the less.

CarloLucibello:
I will merge and ask for someone to test in #201

PR #207, opened 2023-02-11
Adding the UCI Wine dataset

Tambup:
This commit contains the code to build the dataset, a .csv file from https://gist.github.com/tijptjik/9408623, the documentation and test. Everything is based on the current work present in the master branch.

I think this can be useful as it contains more than just two classes but remains quite easy to solve. So, useful to test non-binary classifiers.

I built the documentation locally and it works as expected.
Manual tests has been conducted with success.
Unit Tests has not been concluded (my machine runs out of memory...).


src/datasets/misc/wine.jl
```diff
@@ -0,0 +1,86 @@
+export Wine
+
+"""
+    Wine(; as_df = true, dir = nothing)
+
+The UCI Wine dataset.
```
CarloLucibello, line 6: can you add a reference to the original data source or paper?

Tambup: Done!

Also, I take advantage of this to say that I intentionally avoided to insert the getproperty method as I've seen it is deprecated.  If it should be inserted for compatibility just let me know.

Anyway, great job! I saved a lot of time yesterday!

CarloLucibello: No need for the deprecated methods, I actually they have been around for long time, we should fist remove them

PR #228, opened 2024-04-29
added the PPI dataset

scratch-er:
In this pull request, we add the PPI dataset from http://snap.stanford.edu/graphsage/#datasets

The PPI dataset is loaded in the same way as Reddit. We added a new file `graphsage.jl`, it provides the function `load_graphsage_data`. Most of the code in the original `reddit.jl` is copied into this function, and we added `labels = stack(labels)` to covert the labels of the PPI dataset from `Vector{JSON3.Array}` to a matrix.

 We added `ppi.jl`, which defines the PPI dataset and loads data with this function. Now `reddit.jl` uses this function to load data as well.

### DIFF
--- a/src/MLDatasets.jl
+++ b/src/MLDatasets.jl
@@ -110,6 +110,7 @@ include("graph.jl")
 
 include("datasets/graphs/planetoid.jl")
 include("datasets/graphs/traffic.jl")
+include("datasets/graphs/graphsage.jl")
 # export read_planetoid_data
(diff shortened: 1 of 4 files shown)


CarloLucibello:
Hi, thanks for the contribution and sorry for getting to you so late. This is a welcome addition. You just need to add the new dataset to the documentation, `docs/src/datasets/graph.md`

Issue #239, opened 2024-08-01
Datasets that can be implemented


rbSparky:
This is in comparison to pytorch geometric(some datasets might need to be re verified since they might be implemented in umbrella datasets like OGB):

- [x] Chickenpox
- [x] CiteSeer
- [x] Cora
- [x] KarateClub
- [x] MetrLA
- [x] MovieLens
- [x] OGBDataset
- [x] OrganicMaterialsDB
- [x] PemsBay
- [x] Planetoid
- [x] PolBlogs
- [x] Pubmed
- [x] Reddit
- [x] TemporalBrains
- [x] Traffic
- [x] TuDataset
- [x] WindmillEnergy
- [ ] GNNBenchmarkDataset
- [ ] NELL
- [ ] CitationFull
- [ ] CoraFull
- [ ] Coauthor
- [ ] Amazon
- [ ] Reddit2
- [ ] Flickr
- [ ] Yelp
- [ ] AmazonProducts
- [ ] QM7b
- [x] QM9
- [ ] MD17
- [ ] ZINC
- [ ] AQSOL
- [ ] MoleculeNet
- [ ] PCQM4Mv2
- [ ] Entities
- [ ] RelLinkPredDataset
- [ ] GEDDataset
- [ ] AttributedGraphDataset
- [ ] MNISTSuperpixels
- [ ] FAUST
- [ ] DynamicFAUST
- [ ] ShapeNet
- [ ] ModelNet
- [ ] CoMA
- [ ] SHREC2016
- [ ] TOSCA
- [ ] PCPNetDataset
- [ ] S3DIS
- [ ] GeometricShapes
- [ ] BitcoinOTC
- [ ] GDELTLite
- [ ] ICEWS18
- [ ] GDELT
- [ ] WILLOWObjectClass
- [ ] PascalVOCKeypoints
- [ ] PascalPF
- [ ] SNAPDataset
- [ ] SuiteSparseMatrixCollection
- [ ] WordNet18
- [ ] WordNet18RR
- [ ] FB15k_237
- [ ] WikiCS
- [ ] WebKB
- [ ] WikipediaNetwork
- [ ] HeterophilousGraphDataset
- [ ] Actor
- [ ] UPFD
- [ ] GitHub
- [ ] FacebookPagePage
- [ ] LastFMAsia
- [ ] DeezerEurope
- [ ] GemsecDeezer
- [ ] Twitch
- [ ] Airports
- [ ] LRGBDataset
[...]

PR #250, opened 2026-03-21
Add ZINC molecular graph dataset

Uneeb808:
This PR adds the ZINC molecular graph dataset.

~250,000 molecular graphs with a penalized logP regression target (y = logP - SAS - cycles). Includes the standard 12k benchmark subset from "Benchmarking Graph Neural Networks" (Dwivedi et al. 2020).

Features:
- Full (~250k graphs) and subset (12k) variants
- train/val/test splits
- Node features: atom type (28 classes)
- Edge features: bond type (4 classes)
- Graph-level regression target (Float32)

Note: Raw data is distributed as PyTorch pickle files which Pickle.jl cannot deserialize directly. I preprocessed these to flat NPZ files and hosted them on my fork's releases. Happy to move to JuliaML releases on merge.

### DIFF
--- a/docs/src/datasets/graphs.md
+++ b/docs/src/datasets/graphs.md
@@ -38,4 +38,5 @@ TUDataset
 WindMillEnergy
 AmazonComputers
 AmazonPhoto
+ZINC
 ```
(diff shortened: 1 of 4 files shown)


Uneeb808:
This PR addresses #239

